PR #8, opened 2021-02-12
Add CI and fix lint / fmt errors

pscott:
This PR adds a CI to the repo in order to ensure proper coding manner and to adhere to the community's standards using `cargo fmt` and `cargo clippy`.


src/ecc.rs
```diff
@@ -54,40 +63,53 @@ pub fn ec_get_pubkey(curve: CurvesId, privkey: &mut cx_ecfp_private_key_t) -> Re
     }
 }
 
-pub type DEREncodedECDSASignature = [u8; 73];
+pub type DerEncodedEcdsaSignature = [u8; 73];
```
yhql, line 66: Doesn't work with the acronyms now :D We could use `EncodedSignature` I guess

src/ecc.rs
```diff
@@ -129,73 +151,67 @@ pub const fn make_bip32_path<const N: usize>(bytes: &[u8]) -> [u32; N] {
     let mut i = 2; // parsed character index
     let mut j = 0; // constructed path number index
     let mut acc = 0; // constructed path number
-    let mut state = BIP32ParserState::FirstDigit;
+    let mut state = Bip32ParserState::FirstDigit;
```
yhql, line 154: Same acronym problem as for DEREncodedECDSASignature here

yhql: OR it could just be OK because it's in the rust guidelines to have non upper case acronyms...

PR #9, opened 2021-02-17
Update the NVM module

pscott:
This PR updates the NVM module:
- Update the code to use standard methods like `.position()`, `.get`, or `.fold()`.
- Update the comments and naming conventions
- Modify some return values to `Option` and `Result` to avoid `panic`ing at runtime as much as possible :)

### DIFF
--- a/src/nvm.rs
+++ b/src/nvm.rs
@@ -42,7 +42,6 @@
 
 use crate::bindings::nvm_write;
 use AtomicStorageElem::{StorageA, StorageB};
-use KeyError::{KeyOutOfRange, SlotIsFree};
 
 // Warning: currently alignment is fixed by magic values everywhere, since
 // rust does not allow using a constant in repr(align(...))
@@ -235,11 +234,7 @@ impl<T> SingleStorage<T> for AtomicStorage<T> where T: Copy {
         }
     }
 }
-
-pub enum KeyError {
-    KeyOutOfRange,
-    SlotIsFree,
-}
+pub struct KeyOutOfRange;
 
 /// A Non-Volatile fixed-size collection of fixed-size items.
 /// Items insertion and deletion are atomic.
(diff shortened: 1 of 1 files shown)


pscott:
cc @kingofpayne

src/nvm.rs
```diff
@@ -269,31 +288,35 @@ impl<T, const N: usize> Collection<T, N> where T: Copy {
...
-        let mut result = 0;
-        for v in self.flags.get_ref() {
-            if *v == STORAGE_VALID {
-                result += 1;
-            }
-        }
-        result
+        self.flags.get_ref()
+            .iter()
+            .fold(0, |acc, &byte| acc + (byte == STORAGE_VALID) as u32) as usize
```
pscott, line 314: I initially had written
```rust
self.flags.get_ref()
    .iter()
    .filter(|&byte| *byte == STORAGE_VALID)
    .count()
```
but after digging on the interwebs it turns out [this weird casting is more performant](https://docs.rs/bytecount/0.6.2/src/bytecount/naive.rs.html#11-13)

src/nvm.rs
```diff
@@ -307,26 +328,36 @@ impl<T, const N: usize> Collection<T, N> where T: Copy {
...
 
-    /// Returns the index of an item in the internal storage, given the index
-    /// in the collection. If index is too big, None is returned.
+    /// Counts the number of allocated slots up until `len`.
+    fn count_allocated(&self, len: usize) -> usize {
+        self.flags.get_ref()
+            .iter()
+            .take(len)
+            .fold(0, |acc, &byte| acc + (byte == STORAGE_VALID) as u32) as usize
+    }
```
pscott, line 337: I initially had written
```rust
self.flags.get_ref()
    .iter()
    .filter(|&byte| *byte == STORAGE_VALID)
    .count()
```
but after digging on the interwebs it turns out [this weird casting is more performant](https://docs.rs/bytecount/0.6.2/src/bytecount/naive.rs.html#11-13)

src/nvm.rs
```diff
@@ -307,26 +328,40 @@ impl<T, const N: usize> Collection<T, N> where T: Copy {
...
     /// * `index` - Index in the collection
     fn index_to_key(&self, index: usize) -> Option<usize> {
-        let mut next = 0;
-        let mut count = 0;
+        // Neat optimization: start by setting `next` to index,
+        // because we know we could not have found `index` allocated slots beforehand.
+        let mut next_key = index;
+        // Now count the number of allocated slots we have found up
+        // until this `index` (without including the slot at `index` itself).
+        let mut count = self.count_allocated(index);
```
pscott, line 351: This is the result of https://github.com/LedgerHQ/ledger-nanos-sdk/pull/9/commits/c8a1c122481856751aff18809b3fe9684a3bd5f2 and https://github.com/LedgerHQ/ledger-nanos-sdk/pull/9/commits/54f36733e0c094218835ca70add131f1ff0a6c4a . We could, of course, revert those comits and keep the code a little less confusing.
The performance incrase on this one is linear, so I think it's a _nice to have_

PR #12, opened 2021-05-03
fix: Use String:from_utf8_lossy to print output in build.rs

DCRichards:
## Description

Currently `println!("{:?}", output.stderr);` will print the bytes as a `[u8,u8]`. Use `String::from_utf8_lossy` to convert to a readable string for better debugging.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -40,7 +40,7 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     f.write_all(&output.stdout).unwrap();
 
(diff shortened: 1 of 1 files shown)


DCRichards:
@yhql is this build failure due to same thing you mentioned here https://github.com/LedgerHQ/ledger-nanos-sdk/pull/13#issuecomment-831772500?

yhql:
Yes, same problem, so we'll have to disregard this for now. Thanks for the PR !

PR #17, opened 2021-08-17
Rustier boot

yhql:
- Move boot code from C to Rust
- Remove initial Try/Catch block
- Remove r9-setting hack (not necessary for password-manager rwpi mode anymore)


yogh333:
Hello, is this PR still relevant ? If not, it should be closed.

PR #18, opened 2021-09-01
Fix incorrect comm.get_data function

yhql:
This PR intends to fix the reception of data through APDUs, which was incorrect until now, and also define some convention for APDUs by fixing the use of a single byte length (Lc)


yhql:
Closed in favor of #19

PR #21, opened 2022-02-15
Force on debug symbols

Ericson2314:
Because of `no_default_flags`, these are not set following the DEBUG
environment variable.


Ericson2314:
We could reimplement the `DEBUG` env var logic, but as the release app is not even ELF, I figured there was little point of sometimes not building debug info.

PR #22, opened 2022-06-01
Unified build

yhql:
This PR allows building for Nano S, Nano X and Nano S Plus, through the use of new custom targets.

Bluetooth Low Energy support (Nano X) is also added.


.github/workflows/rust.yml
```diff
@@ -16,28 +16,42 @@ env:
...
         with:
           release: '9-2019-q4'
       - uses: actions/checkout@v2
-      - name: Install thumbv6 target
+      - name: Install thumbv6 target (Nano S, Nano X)
         uses: actions-rs/toolchain@v1
         with:
           toolchain: nightly
           override: true
           target: thumbv6m-none-eabi
```
aiooss-ledger, line 36: I believe `target` is not required anymore and can be removed. See https://github.com/aiooss-ledger/ledger-nano-piv/blob/main/.github/workflows/rust.yml#L27
This means that next step `Install thumbv8 target (Nano S Plus)` can be removed.

README.md
```diff
@@ -24,42 +24,31 @@ Using rustc nightly builds is recommanded as some unstable features are
 required.
 
 - `rustup default nightly`
-- `rustup target add thumbv6m-none-eabi`
+- `rustup target add thumbv6m-none-eabi` for Nano S and Nano X builds
+- `rustup target add thumbv8m.main-none-eabi` for Nano S+ builds
```
aiooss-ledger, line 28: These two lines need to be maybe changed to `rustup component add rust-src`?

README.md
```diff
@@ -24,42 +24,31 @@ Using rustc nightly builds is recommanded as some unstable features are
...
 - install [Clang](http://releases.llvm.org/download.html).
 - install an [ARM GCC toolchain](https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-rm/downloads)
```
aiooss-ledger, line 30: Maybe add a small note about the fact that we need `arm-none-eabi-newlib`?

From my previous experience, on GNU/Linux distributions one can install:
```bash
# On Ubuntu
sudo apt install clang gcc-arm-none-eabi gcc-multilib

# On Fedora
sudo dnf install clang arm-none-eabi-gcc arm-none-eabi-newlib

# On ArchLinux
sudo pacman -S clang arm-none-eabi-gcc arm-none-eabi-newlib
```

aiooss-ledger (approved):

PR #26, opened 2022-07-03
Import asm to build

adlerjohn:
- Remove `#![feature(asm)]` since it's now stable.
- Import `core::arch::asm` to fix build error.

To be merged into #17


adlerjohn:
Poking @yhql!

PR #33, opened 2022-07-30
Add AsRef<[u8]> and Into<[u8; N]> for ECPublicKey

ryankurte:
howdy, just playing with the improvements in #28 and i discovered i couldn't access the value of the new `ECPublicKey` object...

(also out of interest, how come the corresponding Ed25519 public key is 65 bytes, iirc this should be 64?)

### DIFF
--- a/src/ecc.rs
+++ b/src/ecc.rs
@@ -25,6 +25,7 @@ pub enum CurvesId {
     Invalid,
 }
 
+#[derive(Copy, Clone, PartialEq, Debug)]
 pub enum CxError {
     Carry,
     Locked,
@@ -344,6 +345,20 @@ impl<const P: usize, const TY: char> ECPublicKey<P, TY> {
             keylength: P,
             pubkey: [0u8; P],
         }
+    }
+}
+
+/// Access public key value by reference
+impl<const P: usize, const TY: char> AsRef<[u8]> for ECPublicKey<P, TY> {
+    fn as_ref(&self) -> &[u8] {
+        &self.pubkey
+    }
+}
+
+/// Access public key value by reference
+impl<const P: usize, const TY: char> From<ECPublicKey<P, TY>> for [u8; P] {
+    fn from(p: ECPublicKey<P, TY>) -> Self {
+        p.pubkey
     }
 }
 


yhql:
🤦‍♂️ sorry !
Using `AsRef` is a good idea here (and thanks because this is a pattern I'm not used to yet). Do you have some code for your usage that I could add in the `example/signature.rs`? That way I can keep track of how users actually want to use this code.

About Ed25519 public keys:  when using "generate_pair" from the platform, the key is not compressed, and there's one byte of tag. `cx_edwards_compress_point_no_throw` can be used to get the compressed form

ryankurte:
> Do you have some code for your usage that I could add in the example/signature.rs

ahh, not really... the `AsRef<[u8]>` is useful for encoding into APDUs (or directly into the comms buffer), but they don't really stand alone... something like this would be good too though:

```rust
fn sign_hash(hash: &[u8; 32]) -> ([u8; 32], [u8; 64]) {
    let pri_key = ecc::Ed25519::from_bip32(&BIP32_PATH);
    let pub_key = ...

    let sig = ...
    
    (pub_key.to_bytes(), sig.to_bytes())
}
```

> About Ed25519 public keys: when using "generate_pair" from the platform, the key is not compressed, and there's one byte of tag. cx_edwards_compress_point_no_throw can be used to get the compressed form.

interesting, i understand staying expanded from a performance perspective but imo it'd be better to default to the safer / expected compressed version unless folks specifically request expanded?

> cx_edwards_compress_point_no_throw can be used to get the compressed form

do y'all have documentation around the available syscalls or like, doxygen over the C APIs? i'm finding heaps of great high-level documentation but, not so much, around the underlying components...

yhql:
Ok I see. Anyway it's fine, I'll update the example with a few common constructions later.

From your latest commit: is this some debugging feature only or there are arbitrary operations that you really need to do on the private key directly?
I was hoping it would be possible to keep `.key` in `ECPrivateKey` private so that there's no way to unintentionally leak it somewhere :)

> do y'all have documentation around the available syscalls or like, doxygen over the C APIs? i'm finding heaps of great high-level documentation but, not so much, around the underlying components...

The lowest-level documentation is available in the header files of the `cxng` part of the C SDK, for example https://github.com/LedgerHQ/nanos-secure-sdk/blob/master/lib_cxng/include/lcx_ecfp.h#L604-L625
Otherwise the device behaves generally like its emulator [speculos](https://github.com/LedgerHQ/speculos).

(if you can just add the `Eq` to the `derive`s of `CxError` that should greenlight your PR :) )

ryankurte:
> From your latest commit: is this some debugging feature only or there are arbitrary operations that you really need to do on the private key directly?

fair call, we _do_ return generated private keys sometimes but i think that'll be from another derivation path so probably not necessary here. it was mostly so i could diagnose derivations / can remove it from the PR.

> (if you can just add the Eq to the derives of CxError that should greenlight your PR :) )

will do (both)!

PR #35, opened 2022-08-09
Derive keys with the appropriate curves rather than secp256k1

jonored:


jonored:
This should possibly be marked wip; I haven't checked that this works in a real app yet.

dfordivam:
I tried doing ECPublicKey generation using the Ed25519 curve. Everything works good on speculos and NanoS, but somehow the app crashes on S+.

I have figured out that the crash happens at `cx_ecfp_generate_pair_no_throw` API.
I have even tried adding `cx_ecfp_init_private_key_no_throw` API, though it did not resolve the crash.
   
As the crash happens only on S+, it seems that the issue is perhaps not in nano_sdk.

yhql:
Hi, thanks for these modifications.
I undrestand the current status of `bip32_fill` does not cover all paths, but I think passing `self.curve` to the derivation will not work for curves other than Ed25519 or Secp256k1 and r1.
The SDK can be tested through speculos with a `cargo test --features speculos` and those fail with that modification (I know I should document this, apologies. It can't be run in the CI yet)
My second concern is that the temporary key buffer is not wiped on drop, which is something we'd like to do to meet our security guidance.

deadblackclover:
@yhql Maybe there is a way to make then at least support for what exactly works, not just for Secp256k1?

yhql:
Hello @jonored, I reworked that part a bit to essentially apply your fix and more, and get rid of future potential mishaps in https://github.com/LedgerHQ/ledger-nanos-sdk/pull/62 
Do you mind if I merge the above instead (if it indeed does what you expected) ?

jonored:
@yhql Looks good to me, feel free to drop this one. I'm certainly not going to object to getting jobs done with straightforward traits, I just might not get to making a nice one for a bugfix.

PR #42, opened 2022-09-06
cleanup target detection so TARGET works as well as CARGO_CFG_TARGET_OS

ryankurte:
extends target detection in build.rs to work for both `TARGET` and `CARGO_CFG_TARGET_OS` to work with `RUST_TARGET_PATH` tricks (broken by a recent PR) and adds a couple of exported variables for downstream use.

blocked on / includes #41 because it's easier for me to stack PRs, can split if required

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,7 +1,7 @@
 extern crate cc;
 use std::path::PathBuf;
 use std::process::Command;
-use std::{env, error::Error, fs::File, io::Read};
+use std::{env, error::Error, fs::File, io::Read, str::FromStr};
 
 fn finalize_nanos_configuration(command: &mut cc::Build, bolos_sdk: &String) -> String {
     command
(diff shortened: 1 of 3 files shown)


build.rs
```diff
@@ -182,28 +212,18 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-        target_name => panic!(
-            "invalid target `{}`, expected one of `nanos`, `nanox`, `nanosplus`. Run with `-Z build-std=core --target=./<target name>.json`",
-            target_name
-        ),
-    };
+    // Determine which device we're compiling for
+    let device = Device::detect()?;
+
+    // Export `NANOS_SDK_TARGET` and `NANOS_SDK_ROOT` for consuming crates
+    println!("cargo:rustc-env=TARGET={}", device);
```
Ericson2314, line 219: I don't understand why this non-standard stuff is needed? What is wrong with `target_os = nanos` in CFGs?

ryankurte: maybe there's another way of achieving this but i wanted to be able to read `NANOS_SDK_TARGET` from a build script in another package

Ericson2314: My understanding was that `CARGO_CFG_TARGET_OS`, as the standardized env var, will be set by Cargo in every package, so you can just use that?

ryankurte:
closing as resolved by the addition of `target_os` to the target json files, @yhql it'd still be worth adding the `links = "nanos-sdk"` [manifest key](https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key)

PR #44, opened 2022-10-07
Added exit_app to end tests

deadblackclover:
@yhql Fixed `sdk_test_runner` so that after the tests are done, the application will stop

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -132,6 +132,8 @@ pub fn sdk_test_runner(tests: &[&TestType]) {
         debug_print(name);
         debug_print("\n");
     }
+
+    exit_app(0);
 }
 
 /// This variant of `assert_eq!()` returns an error


yhql:
Hi, `exit_app` is called at the end of the test `main` function here https://github.com/LedgerHQ/ledger-nanos-sdk/blob/master/src/lib.rs#L235 
Did you try to use the test framework from an app perhaps?

deadblackclover:
@yhql Hi, yes, I implemented unit tests and ran them with the `cargo test`. Only adding `exit_app(0)` to `sdk_test_runner` helped

yhql:
Hi, apologies for the delay!
I wanted to reproduce the problem first to see where it blocks, but I couldn't. I modified the rust-app locally with this kind of changes:
```
#![feature(custom_test_frameworks)]
#![reexport_test_harness_main = "test_main"]
#![cfg_attr(test, test_runner(nanos_sdk::sdk_test_runner))]

#[cfg(test)]
#[no_mangle]
fn sample_main() {
    test_main();
    nanos_sdk::exit_app(0);
}
```
without adding `exit_app` to the end of the test runner, and it runs tests succesfully.

Do you have a code sample you could provide that reproduces the problem?

deadblackclover:
@yhql Yes, it really works! Maybe I was using a rather old version of the library.

I think we can close PR

yhql:
Ok thanks !

deadblackclover:
@yhql Hi, I was able to catch the error. Apparently I didn't explain it correctly, I use `custom_test_frameworks` implemented in `ledger-nanos-sdk` in my project. That's why stopping tests doesn't work for me.

It would be logical to put `exit_app(0)` in `sdk_test_runner`. I can create a new PR

PR #47, opened 2022-10-17
Targets: add correct options for compilation on some systems

NicsTr:
On some systems (_eg._, mine), the compiler does not seem to use the `ld.lld` option in the custom target definition. This leads to the linker complaining because it cannot find the needed linker script. A way to fix this is to use simultaneously `ld.lld` and and `ld`.

### DIFF
--- a/nanos.json
+++ b/nanos.json
@@ -16,6 +16,10 @@
       "ld.lld": [
           "-Tnanos_layout.ld",
           "-Tlink.ld"
+      ],
+      "ld": [
+          "-Tnanosplus_layout.ld",
+          "-Tlink.ld"
       ]
   },
   "relocation-model": "ropi",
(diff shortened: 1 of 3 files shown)


nanos.json
```diff
@@ -16,6 +16,10 @@
       "ld.lld": [
           "-Tnanos_layout.ld",
           "-Tlink.ld"
+      ],
+      "ld": [
+          "-Tnanosplus_layout.ld",
```
yhql, line 21: should be nanos_layout.ld

nanox.json
```diff
@@ -15,6 +15,10 @@
       "ld.lld": [
           "-Tnanox_layout.ld",
           "-Tlink.ld"
+      ],
+      "ld": [
+          "-Tnanosplus_layout.ld",
```
yhql, line 20: Should be nanox_layout.ld

PR #64, opened 2023-01-09
Make const init of Comm possible.

Ericson2314:
Also derive Debug on Reply.


kingofpayne:
Is the `Debug` derive of `Reply` really necessary? This may involve extra code inclusion, which we like to avoid on embedded targets.

Ericson2314:
It is not necessary, but it also shouldn't impose any cost with `--gc-sections`.

PR #69, opened 2023-02-13
App loading with more features 

Ericson2314:
1. Relocations are performed, obviating the need for manual `pic` sprinkling.

2. Globals with initial values (vs all-0 globals in BSS) are now initialized to those values.

Contains #68, though if that PR is not wanted this is easy to undo.


yhql:
Thanks for this landmark contribution!
I did merge #68.
Wondering about the use of `link_wrap.sh`, which would not work on Windows/Mac (might be a detail but it is nice to not have to worry about this anymore). This  could be integrated into `cargo-ledger` for platform indepence, what do you think?
(I suspect you don't use `cargo-ledger` though 😄)

Ericson2314:
We actually do use `cargo-ledger`! Yeah I suppose it could be made part of that, good idea.

yhql:
Ok, then this is a 'nice' option for cross-platform building, but would be really nice if that wasn't even necessary. I recall you mentioned trying to have a section containing relocations directly in the linker script does not work @Ericson2314 , because then it's also considered something to relocate and trouble ensues. Do you have a link to anything that mentions this problem? Even if it points into LLVM's source code, I'm ready 😄

dfordivam:
The problem with doing the "link-wrap" phase in the cargo ledger is that the `cargo run` and `cargo test` commands will not be able to work. And I am not sure invoking `cargo-ledger` from `cargo test` would be a good idea.

yhql:
Yes agreed! So I just need to find a way to execute those commands in Windows properly, but having a `linker` field that can be different depending on the host does not appear to be possible.

Your PR is great, I'm just unclear on why we can't have the `.rel.*` sections in their own sections. I didn't find any information that would confirm that this is the intended behaviour, or a blindspot of `lld`.

yogh333:
See #128

PR #72, opened 2023-03-27
Allow deriving the instruction from all 4 metadata fields

Ericson2314:
This makes it easier to not forget to check those fields when using `next_event`. Accidental ignoring of these fields has been flagged by auditors.

I was going to pass the metadata by reference, but the borrow-checker didn't like `T: for<'a> TryFrom<&'a ApduMeta>` even with a `+ 'static`; plus, 4 bytes and a 32-bit pointer are the same size. I would hope all this would get inlined so it wouldn't matter anyways.

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -432,3 +432,22 @@ impl IndexMut<usize> for Comm {
         &mut self.apdu_buffer[idx]
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::assert_eq_err as assert_eq;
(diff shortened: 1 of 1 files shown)


yhql:
Hi and thanks very much for this. I only have a very minor nitpick: could you rename `ApduMeta` to `ApduHeader`? This would match the usual description of the CLA | INS | P1 | P2 (although it usually includes `Lc` as well, but since it's implicit in our case...)

Ericson2314:
Done.

PR #73, opened 2023-03-30
reset apdu state when handling invalid INS internally

yogh333:
Fix to handle bad INS properly in SDK


yhql:
Good catch ! Indeed the C SDK resets everything when sending https://github.com/LedgerHQ/ledger-secure-sdk/blob/master/src/os_io_seproxyhal.c#L1365-L1368
I think resetting should be done https://github.com/LedgerHQ/ledger-nanos-sdk/blob/master/src/io.rs#L154-L156 though

yogh333:
Yes indeed...

PR #87, opened 2023-10-06
Introduce zeroize

cchr-ledger:
Introduce zeroize as a dependency and use it to robustly zero-out secrets on drop instead of using core::hint::black_box.


yhql (approved):

yhql:
Could you bump the version (patch)?

PR #88, opened 2023-10-06
Update ELF metadata 

agrojean-ledger:
Update infos.rs (add more metadatas to match C SDK)

Still missing : 

* APPNAME and APPVERSION : don't know how to fetch them at compile time of the SDK...
* SDK_VERSION and SDK_NAME : fields are set to 'None' but we will give them a proper value when we transition to the new rust SDK with https://github.com/LedgerHQ/secure-sdk-rust


link.ld
```diff
@@ -63,6 +61,18 @@ SECTIONS
     *(.ARM.exidx* .gnu.linkonce.armexidx.*)
     *(.debug_info)
   }
+
+  ledger.target (INFO): { KEEP(*(ledger.target)) }
```
agrojean-ledger, line 65: Removed `.` in front of custom sections for compatibility with C sdk. Also, `.` prefixes are supposed to be reserved for "system" section names.

build.rs
```diff
@@ -283,27 +306,7 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-    // do not forget NATIVE_LITTLE_ENDIAN
-    let s = String::from("NATIVE_LITTLE_ENDIAN");
-    defines.push(s.as_str());
-
-    // Add the defines found in the Makefile.conf.cx to our build command.
-    for define in defines {
-        // scott could use for_each
-        command.define(define, None);
-    }
+    extract_defines_from_cx_makefile(&mut command, &cx_makefile);
```
agrojean-ledger, line 309: Added this for better readability but we can remove it.. Not related to what this PR is supposed to do.

agrojean-ledger:
@yhql I didn't add APPNAME and APPVERSION, do you know how to add them when the rust SDK's build.rs is called ?

src/infos.rs
```diff
@@ -10,3 +30,71 @@ static API_LEVEL: u8 = if cfg!(target_os = "nanos") {
...
+#[link_section = "ledger.api_level"]
+static ELF_API_LEVEL: [u8;4] = if cfg!(target_os = "nanos"){
+    *b"0\n\0\0"
+} else if cfg!(target_os = "nanox") {
+    *b"5\n\0\0"
+} else if cfg!(target_os = "nanosplus") {
+    *b"1\n\0\0"
+} else {
+    *b"255\n"
+};
```
xchapron-ledger, line 44: Would need to be retrieved from the SDK, but this for after the change of architecture I guess?
Also, to be consistent with C sdk, the value for nanos should be : undefined, e.g. section doesn't exist.
and the else value might be the same ? Or a build error cause that's just invalid?

agrojean-ledger: Yes in the new SDK we will grab it from the C SDK, it is already done actually here https://github.com/LedgerHQ/secure-sdk-rust/blob/f78cb5803b47de24ef7b2b0323b8b08ad31f7399/build.rs#L180

src/infos.rs
```diff
@@ -10,3 +30,73 @@ static API_LEVEL: u8 = if cfg!(target_os = "nanos") {
...
+
+#[used]
+#[link_section = "ledger.target"]
+static ELF_TARGET: [u8; 8] = if cfg!(target_os = "nanos") {
+    *b"nanos\n\0\0"
```
xchapron-ledger, line 49: Are there solutions to avoid these trailing "\n\0"? It seems a bit weird having them everywhere as I guess htey won't be used by the tools reading elf sections?

yogh333: So I propose to keep it as it is currently (until better proposal)

src/lib.rs
```diff
@@ -5,6 +5,7 @@
 #![test_runner(testing::sdk_test_runner)]
 #![allow(incomplete_features)]
 #![feature(generic_const_exprs)]
+#![feature(const_cmp)]
```
yogh333, line 8: Did not find this nighlty feature in [Unstable Book](https://doc.rust-lang.org/beta/unstable-book/the-unstable-book.html).

1. How did you find it ?
2. Where is it used ?

agrojean-ledger: @yogh333 it's for `let min_length = core::cmp::min(bytes.len(), MAX_METADATA_STRING_SIZE - 1);`

agrojean-ledger: But with @yhql 's proposed solution for strings, we probably would not need it.

agrojean-ledger: And for how I found it, I can't remember ^^

yhql: We can remove this if we replace this `min` by a more manual version:
```
   let min_length = if bytes.len() <= MAX_METADATA_STRING_SIZE - 1 {
        bytes.len()
    } else {
        MAX_METADATA_STRING_SIZE - 1
    };
```

src/infos.rs
```diff
@@ -10,3 +30,73 @@ static API_LEVEL: u8 = if cfg!(target_os = "nanos") {
...
+#[link_section = "ledger.api_level"]
+static ELF_API_LEVEL: [u8; 4] = if cfg!(target_os = "nanos") {
+    *b"0\n\0\0"
+} else if cfg!(target_os = "nanox") {
+    *b"5\n\0\0"
+} else if cfg!(target_os = "nanosplus") {
+    *b"1\n\0\0"
+} else {
+    *b"255\n"
+};
```
xchapron-ledger, line 44: Should be update now that API_LEVEL is now at build time from SDK?

src/infos.rs
```diff
@@ -1,5 +1,25 @@
...
+        }
+        result[count as usize] = bytes[count as usize];
+        count += 1;
+    }
+    result
+}
+
+/// Expose the API_LEVEL to the app
 #[used]
 static API_LEVEL: u8 = if cfg!(target_os = "nanos") {
```
xchapron-ledger, line 24: Should be update to with proper value knwon at build time ?

src/infos.rs
```diff
@@ -10,3 +30,73 @@ static API_LEVEL: u8 = if cfg!(target_os = "nanos") {
...
+    *b"TARGET_NANOX\n\0"
+} else if cfg!(target_os = "nanosplus") {
+    *b"TARGET_NANOS2\n"
+} else {
+    *b"WRONG_TARGET\n\0"
+};
+
+#[used]
+#[link_section = "ledger.sdk_hash"]
+static ELF_SDK_HASH: [u8; 5] = *b"None\n";
```
xchapron-ledger, line 84: I think we can live with this, but best would be having the hash of the C SDK used.

src/infos.rs
```diff
@@ -10,3 +30,73 @@ static API_LEVEL: u8 = if cfg!(target_os = "nanos") {
...
+
+#[used]
+#[link_section = "ledger.sdk_name"]
+static ELF_SDK_NAME: [u8; 18] = *b"ledger-secure-sdk\n";
```
xchapron-ledger, line 88: in case of a nano, with should probably set `nanos-secure-sdk` as it doesn't use the unified one?

src/infos.rs
```diff
@@ -10,3 +30,73 @@ static API_LEVEL: u8 = if cfg!(target_os = "nanos") {
...
+
+#[used]
+#[link_section = "ledger.sdk_version"]
+static ELF_SDK_VERSION: [u8; 5] = *b"None\n";
```
xchapron-ledger, line 92: Same as for the hash, would be best to have the real C SDK version, else that is not a drama.

xchapron-ledger (commented):
I think some stuff could be up^date now that we use ledger_secure_sdk_sys crate.
We could use the env declared here: https://github.com/LedgerHQ/ledger-device-rust-sdk/blob/master/ledger_secure_sdk_sys/build.rs#L208-L210
And maybe add some more to populate properly SDK_HASH / SDK_NAME and SDK_VERSION ?

src/infos.rs
xchapron-ledger: I'm wondering how these trailing \n and \0 will affect our tools reading metadata? Mostly Speculos and LedgerBlue for now.

src/infos.rs
agrojean-ledger: Done in https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/96

src/infos.rs
agrojean-ledger: Done in https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/96

src/infos.rs
agrojean-ledger: Done in https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/96

src/lib.rs
agrojean-ledger: Removed in https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/96

src/infos.rs
agrojean-ledger: Done in https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/96

build.rs
agrojean-ledger: Removed in https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/96

src/infos.rs
agrojean-ledger: Done in https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/96

src/infos.rs
agrojean-ledger: Done in https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/96

agrojean-ledger:
Replaced by https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/96

PR #90, opened 2023-10-11
Use ledger-sdk-sys bindings crate

yhql:
Move C-SDK related build into its own (more or less `-sys`) crate


xchapron-ledger (commented):
there should be an update of version I guess?

PR #95, opened 2023-11-15
Add new UI gadgets

agrojean-ledger:
Add two new UI gadgets : 

* `MultiPageMenu` to display menus with pages that can have various styles : 
    - PictureNormal : Picture (should be 16x16) with two lines of text (page layout depends on device).
    - PictureBold : Icon on top with one line of text on the bottom.
    - BoldNormal : One line of bold text and one line of normal text on Nano S or up to 3 lines on Nano X/SP.
    - Normal : 2 lines of centered text.
 
  `MultiPageMenu` `new` function takes an `io::comm` as parameter so its `show` function can return a page index (in case of user "both button press") or any other event upon reception. It allows for any APDU to be received and processed when a menu is displayed.

* `MultiFieldReview` to review multiple fields with a review message at the start and validation/cancel messages at the end. The `show` function returns a boolean upon validation/cancellation by the user.

### DIFF
--- a/ledger_device_ui_sdk/Cargo.toml
+++ b/ledger_device_ui_sdk/Cargo.toml
@@ -7,9 +7,9 @@ license.workspace = true
 description = "Ledger devices abstractions for displaying text, icons, menus and other common gadgets to the screen"
 
 [dependencies]
-ledger_secure_sdk_sys = "1.0.0"
-ledger_device_sdk = "1.0.0"
-include_gif = "1.0.0"
+ledger_device_sdk = { git = "https://github.com/LedgerHQ/ledger-device-rust-sdk.git", branch = "add-ui-gadgets" }
+ledger_secure_sdk_sys = { git = "https://github.com/LedgerHQ/ledger-device-rust-sdk.git", branch = "add-ui-gadgets" }
+include_gif = { git = "https://github.com/LedgerHQ/ledger-device-rust-sdk.git", branch = "add-ui-gadgets" }
 numtoa = "0.2.4"
 
 [features]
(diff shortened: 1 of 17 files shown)


ledger_device_ui_sdk/Cargo.toml
```diff
@@ -7,8 +7,10 @@ license.workspace = true
 description = "Ledger devices abstractions for displaying text, icons, menus and other common gadgets to the screen"
 
 [dependencies]
-ledger_secure_sdk_sys = "1.0.0"
-include_gif = "1.0.0"
+ledger_device_sdk = { git = "https://github.com/LedgerHQ/ledger-device-rust-sdk.git", branch = "add-ui-gadgets" }
+ledger_secure_sdk_sys = { git = "https://github.com/LedgerHQ/ledger-device-rust-sdk.git", branch = "add-ui-gadgets" }
+include_gif = { git = "https://github.com/LedgerHQ/ledger-device-rust-sdk.git", branch = "add-ui-gadgets" }
+numtoa = "0.2.4"
```
yogh333, line 13: dependencies should be udpated to crates.io :
```
ledger_device_sdk = "1.0.0"
ledger_secure_sdk_sys = "1.0.0"
include_gif = "1.0.0"
```
and new `ledger_device_ui_sdk` should be published on crates.io

I still think that making `ledger_device_ui_sdk` depending on `ledger_device_sdk` is weird and break the planned architecture (even if I understand the needs). That's why I think that later, this crate shall be integrated into `ledger_device_sdk` (as a `ui` module for instance).

yogh333 (approved):

ledger_device_ui_sdk/Cargo.toml
agrojean-ledger: Done

PR #96, opened 2023-11-22
New elf metadata from C SDK.

agrojean-ledger:
Dynamically retrieve C SDK infos and store them in the ELF


ledger_secure_sdk_sys/build.rs
```diff
@@ -71,48 +71,212 @@ const CCID_FILES: [&str; 9] = [
...
+    pub target_id: String,
+    pub target_name: String,
+    pub c_sdk_name: String,
+    pub c_sdk_hash: String,
+    pub c_sdk_version: String,
+}
+
+impl SDKInfo {
+    pub fn new() -> Self {
+        SDKInfo {
```
yhql, line 104: you can add a `#[derive(Default)]` above `struct SDKInfo` line 92, then simply have
```
pub fn new() -> Self {
    SDKInfo::default()
}
```

ledger_secure_sdk_sys/build.rs
```diff
@@ -71,48 +71,212 @@ const CCID_FILES: [&str; 9] = [
...
+    };
+    let target_file_path = bolos_sdk.join(format!("{}include/bolos_target.h", prefix));
+    println!("cargo:warning=Target file path is {:?}", target_file_path);
+    let target_file =
+        File::open(target_file_path).map_err(|_| SDKBuildError::TargetFileNotFound)?;
+    let mut target_id = String::new();
+    let mut target_name = String::new();
+    for line in BufReader::new(target_file).lines().flatten() {
+        if line.contains("TARGET_ID") {
+            let id = line.split(" ").collect::<Vec<&str>>()[2];
```
yhql, line 215: `let id = line.split(" ").nth(2).unwrap();`

ledger_secure_sdk_sys/build.rs
```diff
@@ -71,48 +71,212 @@ const CCID_FILES: [&str; 9] = [
...
+            "https://github.com/LedgerHQ/nanos-secure-sdk",
+            "master",
+            0,
+            "nanos-secure-sdk",
+        ),
         Device::NanoX => (
             "https://github.com/LedgerHQ/ledger-secure-sdk",
             "API_LEVEL_5",
             5,
+            "ledger-secure-sdk",
```
yhql, line 244: the `name` could be retrieved by making the repo_url a PathBuf and using https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.file_name

ledger_secure_sdk_sys/build.rs
agrojean-ledger: Done.

ledger_secure_sdk_sys/build.rs
agrojean-ledger: Updated the way we iterate to make sure we only parse the values once.

ledger_secure_sdk_sys/build.rs
agrojean-ledger: Ok

ledger_secure_sdk_sys/build.rs
```diff
@@ -71,48 +71,217 @@ const CCID_FILES: [&str; 9] = [
...
+            .arg(sdk_info.bolos_sdk.as_path())
             .output()
             .ok();
     }
-    (bolos_sdk, api_level)
+    sdk_info.api_level = api_level;
+    (sdk_info.target_id, sdk_info.target_name) =
+        retrieve_target_file_infos(device, &sdk_info.bolos_sdk)?;
+    (sdk_info.c_sdk_hash, sdk_info.c_sdk_version) = retrieve_sdk_git_info(&sdk_info.bolos_sdk);
+    sdk_info.c_sdk_name = repo_url.file_name().unwrap().to_str().unwrap().to_string();
```
xchapron-ledger, line 274: You could `retrieve_sdk_info` here I guess?
And I even think that we could do it directly in `pub fn bolos_sdk` either with the path to the already present SDK, or with the path of the cloned SDK (path that would be the only return of `clone_sdk`

ledger_secure_sdk_sys/build.rs
```diff
@@ -71,48 +71,217 @@ const CCID_FILES: [&str; 9] = [
...
+    let makefile_defines =
+        File::open(bolos_sdk.join("Makefile.defines")).expect("Could not find Makefile.defines");
+    let mut api_level: Option<u32> = None;
+    let mut sdk_name: Option<String> = None;
+    for line in BufReader::new(makefile_defines).lines().flatten() {
+        if let Some(value) = line.split(":=").nth(1).map(str::trim) {
+            if line.contains("API_LEVEL") && api_level.is_none() {
+                api_level = Some(value.parse().map_err(|_| SDKBuildError::InvalidAPILevel)?);
+            } else if line.contains("SDK_NAME") && sdk_name.is_none() {
+                sdk_name = Some(value.to_string());
```
xchapron-ledger, line 176: the value here contains `"` that should be removed.

ledger_secure_sdk_sys/build.rs
```diff
@@ -173,42 +342,37 @@ impl SDKBuilder {
...
+            format!("{}", self.api_level)
+        );
         println!("cargo:warning=API_LEVEL is {}", self.api_level);
-
+        println!("cargo:rustc-env=TARGET_ID={}", sdk_info.target_id);
+        println!("cargo:warning=TARGET_ID is {}", sdk_info.target_id);
+        println!("cargo:rustc-env=TARGET_NAME={}", sdk_info.target_name);
+        println!("cargo:warning=TARGET_NAME is {}", sdk_info.target_name);
+        println!("cargo:rustc-env=C_SDK_NAME={}", sdk_info.c_sdk_name);
+        println!("cargo:warning=C_SDK_NAME is {}", sdk_info.c_sdk_name);
```
xchapron-ledger, line 371: In the deployer script, we use only the sdk_name field, where we put `"Rust SDK + " "ledger-secure-sdk"`. It ends in the yml artifact and in the tag pushed on the repo.
I think there is no great reason to have a different behavior, though I'm not sure which version we should choose.

xchapron-ledger (commented):
Tested on app build and Speculos, it works well!

Do not forget to bump the crate versions.

ledger_secure_sdk_sys/build.rs
agrojean-ledger: Done.

ledger_secure_sdk_sys/build.rs
agrojean-ledger: Done.

ledger_secure_sdk_sys/build.rs
```diff
@@ -71,26 +71,183 @@ const CCID_FILES: [&str; 9] = [
...
+            }
+        }
+
+        if api_level.is_some() && sdk_name.is_some() {
+            // Both keys found, break out of the loop
+            break;
+        }
+    }
+
+    let api_level = api_level.ok_or(SDKBuildError::InvalidAPILevel)?;
```
xchapron-ledger, line 186: This is legitimate on nanos SDK. On C apps it default to no define nor metadata section.

agrojean-ledger: I removed the field for Nano S

ledger_secure_sdk_sys/src/infos.rs
```diff
@@ -1,3 +1,22 @@
...
+
+macro_rules! const_cstr {
+    ($name: ident, $section: literal, $in_str: expr) => {
+        #[used]
+        #[link_section = $section]
+        static $name: [u8; $in_str.len() + 1] = make_c_string($in_str);
+    };
+}
+
 const fn const_parse_api_level(x: &str) -> u8 {
```
xchapron-ledger, line 20: Need `#[cfg(not(target_os = "nanos"))]` to avoid build warning 

```
function `const_parse_api_level` is never usedr_secure_sdk_sys                                                                                                                                           
  --> /home/xchapron/Workspace/LedgerHQ/ledger-device-rust-sdk/ledger_secure_sdk_sys/src/infos.rs:20:10
```

xchapron-ledger (commented):
Tested ok on Speculos on all devices.

ledger_secure_sdk_sys/build.rs
```diff
@@ -71,26 +71,180 @@ const CCID_FILES: [&str; 9] = [
...
+}
+
+fn retrieve_makefile_infos(bolos_sdk: &Path) -> Result<(Option<u32>, String), SDKBuildError> {
+    let makefile_defines =
+        File::open(bolos_sdk.join("Makefile.defines")).expect("Could not find Makefile.defines");
+    let mut api_level: Option<u32> = None;
+    let mut sdk_name: Option<String> = None;
+    for line in BufReader::new(makefile_defines).lines().flatten() {
+        if let Some(value) = line.split(":=").nth(1).map(str::trim) {
+            if line.contains("API_LEVEL") && api_level.is_none() {
```
xchapron-ledger, line 173:
```suggestion
            if line.contains("API_LEVEL") && api_level.is_none() {
```
Does fmt check trailing whitespaces?

xchapron-ledger: ok CI says it does :)

ledger_secure_sdk_sys/src/infos.rs
agrojean-ledger: Done

ledger_secure_sdk_sys/build.rs
agrojean-ledger: I pushed the formatted code

PR #97, opened 2023-11-28
Improvements in commands parsing support

kingofpayne:
Translation from an `ApduHeader` to a custom command type can now return different status words, which are then automatically sent as reply to the host by the SDK.

This allows to implement complex parsing logic which may return appropriate error status words depending on which APDU header fields are invalid (invalid class, invalid instruction, bad P1 or P2, etc.).

Added a check to reject automatically too short APDUs, so SDK users don't have to do this check in the app code.

### DIFF
--- a/ledger_device_sdk/Cargo.toml
+++ b/ledger_device_sdk/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ledger_device_sdk"
-version = "1.1.0"
+version = "1.2.0"
 authors = ["yhql", "yogh333"]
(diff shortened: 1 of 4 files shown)


agrojean-ledger:
@kingofpayne this needs to be rebased following the merge of https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/98

yhql (approved):

xchapron-ledger (commented):
Please do not forget to bump the crate version.

PR #98, opened 2023-11-30
Feat/merge UI sdk

yogh333:
Merged `ui` crate into main SDK


ledger_secure_sdk_sys/src/c/src.c
```diff
@@ -28,7 +28,7 @@ int c_main(void) {
     BEGIN_TRY {
       TRY {
         // below is a 'manual' implementation of `io_seproxyhal_init`
-        check_api_level(CX_COMPAT_APILEVEL);
+        // check_api_level(CX_COMPAT_APILEVEL);
```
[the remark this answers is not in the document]
yogh333: yes indeed

xchapron-ledger (commented):
I'm wondering if we shouldn't make it to separate PR?
One on the sys crate, doing what necessary and bumping the version.
then merging this one and deploying it.
So that we can create another on on the merge of the 2 other crate without having to tweak the dependencies version to make the CI pass.

yogh333:
Not sure about why you are talking about 2 PRs: modifications performed in the sys crate are only due to `Clippy` or `cargo fmt`. There're will later heavy changes when I will do the modifications linked with the bindings that we talked about. And specifying multiple dependencies is common and perfectly supported: [see here](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations).

ledger_device_sdk/Cargo.toml
```diff
@@ -14,10 +14,12 @@ ledger_device_sdk = { path = ".", features = ["speculos"] }
 testmacro = { git = "https://github.com/yhql/testmacro"}
 
 [dependencies]
-ledger_secure_sdk_sys = "1.0.1" 
+ledger_secure_sdk_sys = {path = "../ledger_secure_sdk_sys", version = "1.0.2"}
+include_gif = {path = "../include_gif", version = "1.0.1"}
```
xchapron-ledger, line 18: I'm talking about this, shouldn't we use the version from crates.io?

yogh333: no, I will publish a new version of `include_gif` on `crates.io` once the PR will be merged, it will be version 1.0.1.

yogh333: Even if it is true that the modification in this crate is only relevant when building from a workspace environment. But it should not introduce any regression...

yogh333: same for `ledger_secure_sdk_sys`,  new version will be published on crates.io as soon as the PR is merged

yogh333: And last but not least, version is always checked, even locally, so even when working with local path, it is the same version as the one pulled from crates.io

agrojean-ledger: Why not publish before merging with proper versions ?

agrojean-ledger: @xchapron-ledger indeed, didn't know about that one. Cool stuff. Let's go.

agrojean-ledger (approved):

PR #103, opened 2023-12-08
Feat/add cargo ledger

yogh333:


.github/workflows/ci.yml
```diff
@@ -22,15 +22,31 @@ jobs:
...
         with:
           command: clippy
-          args: --target ${{ matrix.target }}
+          args: -p ${{ matrix.package }} --target ${{ matrix.target }}
   
+  clippy-cargo-ledger:
+    name: Run static analysis for cargo-ledger
+    runs-on: ubuntu-latest
+    container:
+        image: ghcr.io/ledgerhq/ledger-app-builder/ledger-app-dev-tools:latest
```
agrojean-ledger, line 39: Not sure we need to be in the dev-tools container for cargo-ledger jobs... :thinking:

yogh333: Well, the idea is to be consistent with regards to other CI tests: all are performed within dev-tools container, so I have decided to keep the same test environment. For cargo-ledger, we are using a stable toolchain for cargo-ledger, so there's should not be any diff even if we were using an ubuntu-based only container.

cargo-ledger/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "cargo-ledger"
+version = "1.2.0"
```
agrojean-ledger, line 3: Bump the version ?

agrojean-ledger: Then we should update the version in the app-builder docker images

yogh333: OK I will bump the version to 1.2.1

agrojean-ledger: I will update the version in the app builder image in https://github.com/LedgerHQ/ledger-app-builder/pull/91

PR #105, opened 2023-12-14
ui:gadgets.rs: Add display_pending_review()

xchapron-ledger:
Linked to https://github.com/LedgerHQ/app-boilerplate-rust/pull/41

### DIFF
--- a/ledger_device_sdk/src/ui/gadgets.rs
+++ b/ledger_device_sdk/src/ui/gadgets.rs
@@ -4,14 +4,14 @@ use core::str::from_utf8;
 
 use crate::{
     buttons::ButtonEvent::*,
-    io::{self, ApduHeader, Reply},
+    io::{self, ApduHeader, Comm, Event, Reply},
 };
 use ledger_secure_sdk_sys::{
     buttons::{get_button_event, ButtonEvent, ButtonsState},
     seph,
 };
 
-use crate::ui::bitmaps::Glyph;
+use crate::ui::bitmaps::{Glyph, WARNING};
 
 use crate::ui::{bagls::*, fonts::OPEN_SANS};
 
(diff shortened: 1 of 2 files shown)


ledger_device_sdk/src/ui/gadgets.rs
```diff
@@ -66,6 +66,31 @@ pub fn clear_screen() {
     BLANK.paint();
 }
 
+// Developer mode / pending review popup
+// must be cleared with user interaction
+pub fn display_pending_review(comm: &mut Comm) {
```
agrojean-ledger, line 71: Are we sure we want this function defined in the gadgets module ?

xchapron-ledger: Where do you think it would fit better?

kingofpayne: Aside the `popup` method seems the most logical at the moment.

ledger_device_sdk/src/ui/gadgets.rs
```diff
@@ -66,6 +66,51 @@ pub fn clear_screen() {
...
+///     ...
+/// }
+/// `
+pub fn display_pending_review(comm: &mut Comm) {
+    clear_screen();
+
+    // Add icon and text to match the C SDK equivalent.
+    if cfg!(target_os = "nanos") {
+        "Pending".place(Location::Custom(2), Layout::Centered, true);
+        "Ledger review".place(Location::Custom(14), Layout::Centered, true);
```
yhql, line 97: Something like
```
[ "Pending", "Ledger review" ].place(Location::Middle, Layout::Centered, true);
```
should work also?

ledger_device_sdk/src/ui/gadgets.rs
```diff
@@ -66,6 +66,51 @@ pub fn clear_screen() {
...
+    } else {
+        WARNING.draw(57, 10);
+        "Pending".place(Location::Custom(28), Layout::Centered, true);
+        "Ledger review".place(Location::Custom(42), Layout::Centered, true);
+    }
```
kingofpayne, line 102: `Page::from((["Pending", "Review"], &WARNING)).place();` could almost do the same (except bold text). Should we try to use this instead?

kingofpayne: `Page::new(PageStyle::PictureBold, ["Pending", "Review"], Some(&WARNING)).place();`

Should work, but there is a minor bug in the `Page` implementation on the `PictureBold` case that should be fixed at the same time (drawing can almost be copied from the `PictureNormal` case).

ledger_device_sdk/src/ui/gadgets.rs
xchapron-ledger: I doubt this would make it pixel perfect with the C SDK version, and that's what I wanted.

ledger_device_sdk/src/ui/gadgets.rs
xchapron-ledger: I doubt this would make it pixel perfect with the C SDK version, and that's what I wanted.

kingofpayne (approved):
I believe the code can be simplified and reuse the `Page` struct of the SDK, as stated in the comments. However for the moment `Page` has an issue with bold text that would need to be solved first.
So in order to progress, I think we can merge this, and improve the code later when it will be possible.

PR #107, opened 2023-12-15
Added automatic CLA filtering option in Comm

kingofpayne:
Most of the applications check for the value of CLA byte of incoming APDUs and return an error status word when it is different from a constant.

This feature can help simplifying apps by removing the need to handle bad CLA errors, as there is not added value by the apps in dealing with this.

### DIFF
--- a/ledger_device_sdk/Cargo.toml
+++ b/ledger_device_sdk/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ledger_device_sdk"
-version = "1.3.0"
+version = "1.4.0"
 authors = ["yhql", "yogh333"]
 edition = "2021"
 license.workspace = true
--- a/ledger_device_sdk/Cargo.toml
+++ b/ledger_device_sdk/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ledger_device_sdk"
-version = "1.2.1"
+version = "1.3.0"
 authors = ["yhql", "yogh333"]
 edition = "2021"
 license.workspace = true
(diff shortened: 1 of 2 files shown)


agrojean-ledger:
@kingofpayne you need to rebase your branch

ledger_device_sdk/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ledger_device_sdk"
-version = "1.2.1"
+version = "1.3.0"
```
agrojean-ledger, line 3: You need to bump to `1.4.0` (it's already `1.3.0` on master)

ledger_device_sdk/src/io.rs
```diff
@@ -129,15 +136,36 @@ pub struct ApduHeader {
...
+    /// * `cla` - Expected value for APDUs CLA byte.
+    ///
+    /// # Examples
+    ///
+    /// This method can be used when building an instance of [`Comm`]:
+    ///
+    /// ```
+    /// let mut comm = Comm::new().expect_cla(0xe0);
+    /// ```
+    pub fn expect_cla(mut self, cla: u8) -> Self {
```
agrojean-ledger, line 164: I would rather call this function `set_expected_cla`

ledger_device_sdk/Cargo.toml
kingofpayne: Done.

ledger_device_sdk/src/io.rs
kingofpayne: Good suggestion. Done.

agrojean-ledger (approved):

PR #109, opened 2023-12-18
Feat/stax nbgl

agrojean-ledger:
WIP


yogh333:
Seems to be a dupplicate of #114. Should be closed ?

agrojean-ledger:
> Seems to be a duplicate of #114. Should be closed ?

This PR is from the original "proof of concept" branch from @yhql. We can close it indeed, it is redundant (and not as up to date) as #114.

PR #110, opened 2023-12-18
feat: expose API for EIP2333 key derivation

bochaco:


ledger_device_sdk/src/ecc.rs
```diff
@@ -403,6 +403,19 @@ pub fn bip32_derive(curve: CurvesId, path: &[u32], key: &mut [u8]) -> Result<(),
     Ok(())
 }
 
+/// Wrapper for 'os_perso_derive_eip2333'
+pub fn eip2333_derive(path: &[u32], key: &mut [u8]) -> Result<(), CxError> {
+    unsafe {
+        os_perso_derive_eip2333(
```
xchapron-ledger, line 409: I know it's what we have done already, but shouldn't we change it to use the no_throw version of the syscalls / cx_calls?

ledger_device_sdk/src/ecc.rs
```diff
@@ -495,6 +508,19 @@ impl SeedDerive for Stark256 {
     }
 }
 
+impl SeedDerive for Bls12381G1 {
+    type Target = ECPrivateKey<32, 'E'>;
```
yhql, line 512: Bls12381G1 is a Weierstrass curve, so should be `ECPrivateKey<32, 'W'>` ideally

ledger_device_sdk/src/ecc.rs
```diff
@@ -530,6 +556,7 @@ impl_curve!(BrainpoolP512R1, 64, 'W');
 impl_curve!(BrainpoolP512T1, 64, 'W');
 impl_curve!(Stark256, 32, 'W');
 impl_curve!(Ed25519, 32, 'E');
+impl_curve!(Bls12381G1, 32, 'E'); // TODO: review type of curve (...and sigining impl)
```
yhql, line 559: I guess we should not do this and expose the usual signing APIs yet. EdDSA or ECDSA are typically not used on BLS12381G1. We need to find a way to expose signing on BLS12381G2 though, which is a BLS signature, not ECDSA or EdDSA, and whose keys must be derived with eip2333 using G1

ledger_device_sdk/src/ecc.rs
xchapron-ledger: I'm not sure, they are static inline function defines in os_perso.h @yhql  maybe it can't be used to generate bindings?

ledger_device_sdk/src/ecc.rs
bochaco: Agreed, apart from impl. details w.r.t. being able to use the `impl_curve!` or not that I'll try to work out, I've found there is `ox_bls12381_sign`, is that supposed to be good enough for BLS12381G2 sig?

yogh333:
Could we rename this PR into "BLS12-381 support" ?

PR #114, opened 2024-01-10
Implement POC of Stax app with home + settings screen

agrojean-ledger:


ledger_device_sdk/src/io.rs
```diff
@@ -268,7 +280,80 @@ impl Comm {
...
+        //     let event: Option<Event<T>> = self.check_event();
+        //     if event.is_some() {
+        //         return true;
+        //     }
+        // }
+
+        let mut spi_buffer = [0u8; 128];
+
+        // unsafe {
+        //     G_io_app.apdu_state = APDU_IDLE;
```
agrojean-ledger, line 298: Calling this each time we enter the `next_event_ahead` function messes up the IO global state and results in the reply not being sent for instance after signing a transaction.

Maybe we should call it once in an init function before looping on `next_event_ahead` ? @kingofpayne

ledger_device_sdk/src/io.rs
```diff
@@ -268,7 +280,80 @@ impl Comm {
         }
     }
 
-    pub fn decode_event<T>(&mut self, spi_buffer: &mut [u8; 128]) -> Option<Event<T>>
+    pub fn next_event_ahead<T>(&mut self) -> bool
+    where
+        T: TryFrom<ApduHeader>,
+        Reply: From<<T as TryFrom<ApduHeader>>::Error>,
+    {
+        // if self.event_pending {
```
agrojean-ledger, line 288: It seems like this is not needed... it works with or without it... Not sure why @kingofpayne

ledger_device_sdk/src/nbgl.rs
```diff
@@ -0,0 +1,267 @@
...
+    Validate,
+    Reject,
+}
+
+pub struct Field<'a> {
+    pub name: &'a str,
+    pub value: &'a str,
+}
+
+struct NbglContext {
```
agrojean-ledger, line 23: This struct should be cleaned-up, most of it, if not all of it probably is not needed with the nbgl sync API and could be replaced by fields of `NbglHome`, `NbglReview` and `NbglAddressConfirm`

ledger_device_sdk/src/nbgl.rs
```diff
@@ -0,0 +1,267 @@
...
+        self
+    }
+
+    pub fn show<T: TryFrom<ApduHeader>>(&mut self) -> Event<T>
+    where
+        Reply: From<<T as TryFrom<ApduHeader>>::Error>,
+    {
+        unsafe {
+            loop {
+                match ledger_secure_sdk_sys::sync_nbgl_useCaseHomeAndSettings() {
```
agrojean-ledger, line 93: A "real" `HomeAndSettings` sync API should be used here, this version was just for proof of concept purposes and is pretty bare bones.

ledger_device_sdk/src/nbgl.rs
```diff
@@ -0,0 +1,267 @@
...
+            let sync_ret = sync_nbgl_useCaseAddressReview(
+                address.as_ptr() as *const ::core::ffi::c_char,
+                &icon as *const nbgl_icon_details_t,
+                "Verify address\0".as_ptr() as *const ::core::ffi::c_char,
+                core::ptr::null(),
+            );
+
+            match sync_ret {
+                ledger_secure_sdk_sys::NBGL_SYNC_RET_SUCCESS => {
+                    return true;
```
agrojean-ledger, line 206: `sync_nbgl_useCaseStatus` should be added upon confirmation / rejection of address

ledger_device_sdk/src/io.rs
kingofpayne: I think the `next_event_ahead` method should not call again `seph_recv` when an APDU is pending. When an APDU is pending, the MCU expects a reply to that event, but if you call again `next_event_ahead` while the response has not been sent I think this is where you get troubles (that could explain why you had to comment out the global state init).

ledger_device_sdk/src/io.rs
kingofpayne: See previous response related to this.
Maybe the successive calls to `seph_recv` silently fail or do nothing because you can't receive a new event as you did not reply to the pending APDU. If this is the case then the apdu buffer remains unmodified and still holds the previous APDU, so `detect_apdu` would still return `true`.

include_gif/src/lib.rs
```diff
@@ -1,25 +1,103 @@
...
+    }
+}
 
 #[proc_macro]
 pub fn include_gif(input: TokenStream) -> TokenStream {
-    let filename = syn::parse_macro_input!(input as syn::LitStr);
+    let input = parse_macro_input!(input as Input);
+    match input {
+        Input::FileNameOnly(filename) => {
+            // Default to Bagl if no type is specified.
```
yogh333, line 50: As far as I have understood, BAGL will be soon replaced by NBGL, even for Nano devices. So maybe default value should be set to NBGL ?

include_gif/src/lib.rs
```diff
@@ -1,25 +1,103 @@
...
+            let packed = generate_bagl_glyph(&frame, &palette);
+            write!(
+                &mut vec_output,
+                "(&{:?}, {}, {})",
+                packed, frame.width, frame.height
+            )
+            .unwrap();
+        }
+        GlyphType::Nbgl => {
+            let (compressed_buffer, bpp) = generate_nbgl_glyph(&frame, &palette);
```
yogh333, line 86: Does it mean that we only support compressed glyphs ? Maybe I am wrong, but it seems to me that some glyphs shall be supported in uncompressed format...

include_gif/src/lib.rs
agrojean-ledger: Risky choice, we should probably discuss that in our weekly meeting. :thinking: 
What do you think @xchapron-ledger ?

include_gif/src/lib.rs
agrojean-ledger: I'm not sure there are benefits to uncompressed glyphs... they just take more memory :man_shrugging: . We could have options, I'm not against it, but not sure it would be really useful :thinking:

include_gif/src/lib.rs
agrojean-ledger: @yogh333 actually it's used in quite a few apps : https://github.com/search?q=org%3ALedgerHQ+include_gif+language%3ARust&type=code&l=Rust

yogh333: Let's discuss about that during our next meeting, but even for existing apps, the modification is minor and pretty straightforward and it could be more future proof to me.

include_gif/src/lib.rs
yogh333: Yes I agree with you regarding the "benefit" of using uncompressed glyphs but I remark this in the C SDK python script: https://github.com/LedgerHQ/ledger-secure-sdk/blob/e55673807f200645d2aacbd87dbfbb514b6df56b/lib_nbgl/tools/icon2glyph.py#L351C16-L351C72, so it may be relevant to consider it maybe 🤔

agrojean-ledger: @yogh333 indeed. Do you prefer we keep the same way of doing in things as in the C SDK script (using a `_nocomp` suffix on file names) or we add a new value for the 2nd parameter of our macro such as 'NBGL_NOCOMP' ? I would maybe prefer the second option. :thinking:

yogh333: ok for the second option

ledger_device_sdk/src/nbgl.rs
yogh333, line 1: Should maybe put into ui module ?

agrojean-ledger: Not really as the ui module is not compatible with Stax or Flex (we don't compile the ui module at all when the target is Stax or Flex), cf `lib.rs`

```
#[cfg(any(target_os = "stax", target_os = "flex"))]
pub mod nbgl;
#[cfg(not(any(target_os = "stax", target_os = "flex")))]
pub mod ui;
```

cargo-ledger/Cargo.toml
```diff
@@ -1,8 +1,8 @@
 [package]
 name = "cargo-ledger"
-version = "1.3.0"
-authors = ["yhql"]
-description = "Build and sideload Ledger Nano apps"
+version = "1.5.0"
```
yogh333, line 3: Why not `1.4.0` ?

ledger_device_sdk/build.rs
```diff
@@ -30,7 +34,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     let linkerscript = match device {
         NanoS => "nanos_layout.ld",
         NanoX => "nanox_layout.ld",
-        NanoSPlus => "nanosplus_layout.ld",
+        NanoSPlus | Stax | Flex => "nanosplus_layout.ld",
```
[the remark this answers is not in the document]
agrojean-ledger: Good catch, I don't really know, it was something I took from @yhql's original PR, which used the same as Nano S Plus for Stax, so I added Flex as well, but we need to make sure it's correct.

cargo-ledger/src/setup.rs
```diff
@@ -13,13 +13,13 @@ pub fn install_targets() {
     let sysroot_cmd = std::str::from_utf8(&sysroot_cmd).unwrap().trim();
 
     let target_files_url = Path::new(
-        "https://raw.githubusercontent.com/LedgerHQ/ledger-device-rust-sdk/a630c93398316cc56dcec6aceb0d424b626d150c/ledger_device_sdk/"
+        "https://raw.githubusercontent.com/LedgerHQ/ledger-device-rust-sdk/feat/stax_nbgl_2/ledger_device_sdk/"
```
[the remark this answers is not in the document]
agrojean-ledger: Indeed ! It was for testing but I'll put back a commit hash.

cargo-ledger/Cargo.toml
agrojean-ledger: Typo, I'll fix this.

ledger_device_sdk/build.rs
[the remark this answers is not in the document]
agrojean-ledger: Done.

cargo-ledger/src/setup.rs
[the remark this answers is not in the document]
agrojean-ledger: Done.

ledger_device_sdk/src/nbgl.rs
agrojean-ledger: Done.

ledger_device_sdk/src/nbgl.rs
agrojean-ledger: Done.

PR #115, opened 2024-01-10
Feat/update nanosplus sdk

yogh333:
- [ledger_secure_sdk_sys] Update C SDK for Nano S+ if cloned when building Rust SDK
- [cargo-ledger] retrieve C SDK from app-builder docker container

### DIFF
--- a/cargo-ledger/src/main.rs
+++ b/cargo-ledger/src/main.rs
@@ -202,9 +202,26 @@ fn build_app(
                     "--message-format=json-diagnostic-rendered-ansi",
                 ])
                 .args(&remaining_args)
-                .stdout(Stdio::piped())
-                .spawn()
-                .unwrap();
+                .stdout(Stdio::piped());
+
+            let c_sdk_path = match (device) {
+                Device::Nanos => std::env::var("NANOS_SDK"),
+                Device::Nanosplus => std::env::var("NANOSP_SDK"),
+                Device::Nanox => std::env::var("NANOX_SDK"),
+            };
+
+            match c_sdk_path {
+                Ok(path) => {
+                    cargo_cmd = cargo_cmd
+                        .arg(format!("--config env.LEDGER_C_SDK_PATH={}", path))
+                }
+                Err(_) => {
+                    println!("No C SDK, it will be cloned while building the Rust SDK");
+                    ()
+                }
+            }
+
+            cargo_cmd.spawn().unwrap();
 
             let mut exe_path = PathBuf::new();
             let out = cargo_cmd.stdout.take().unwrap();
--- a/ledger_secure_sdk_sys/build.rs
+++ b/ledger_secure_sdk_sys/build.rs
@@ -215,7 +215,7 @@ fn clone_sdk(device: &Device) -> PathBuf {
         ),
         Device::NanoSPlus => (
             Path::new("https://github.com/LedgerHQ/ledger-secure-sdk"),
-            "API_LEVEL_1",
+            "API_LEVEL_5",
         ),
     };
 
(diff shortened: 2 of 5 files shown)


cargo-ledger/src/main.rs
```diff
@@ -202,9 +202,25 @@ fn build_app(
...
+            let c_sdk_path = match device {
+                Device::Nanos => std::env::var("NANOS_SDK"),
+                Device::Nanosplus => std::env::var("NANOSP_SDK"),
+                Device::Nanox => std::env::var("NANOX_SDK"),
+            };
+
+            match c_sdk_path {
+                Ok(path) => {
+                    cargo_cmd = cargo_cmd
+                        .arg(format!("--config env.LEDGER_C_SDK_PATH={}", path))
```
xchapron-ledger, line 216: Maybe check if `LEDGER_C_SDK_PATH` is not already set?

ledger_secure_sdk_sys/build.rs
```diff
@@ -330,8 +330,8 @@ impl SDKBuilder {
     }
 
     pub fn bolos_sdk(&mut self) -> Result<(), SDKBuildError> {
-        println!("cargo:rerun-if-env-changed=LEDGER_SDK_PATH");
-        let sdk_path = match env::var("LEDGER_SDK_PATH") {
+        println!("cargo:rerun-if-env-changed=LEDGER_C_SDK_PATH");
+        let sdk_path = match env::var("LEDGER_C_SDK_PATH") {
```
xchapron-ledger, line 334: This is used in deployment script, so if not necessary, would be better not to change it.

cargo-ledger/src/main.rs
xchapron-ledger: Needed for deployement pipeline in fact, as `NANOS_SDK` and others will be set, but we don't want their values.

ledger_secure_sdk_sys/build.rs
yogh333: Ok

cargo-ledger/src/main.rs
yogh333: ok done

cargo-ledger/src/main.rs
```diff
@@ -194,13 +194,34 @@ fn build_app(
...
+
+            match std::env::var("LEDGER_SDK_PATH") {
+                Ok(_) => (),
+                Err(_) => {
+                    if let Ok(path) = c_sdk_path {
+                        args.push(format!(
+                            "--config env.LEDGER_SDK_PATH={}",
+                            path
+                        ))
+                    }
```
yhql, line 219: Should we do something when `c_sdk_path` is an `Err()` too? Abort?

yogh333: Yes in that case, C SDK will be have to be cloned when building sys crate

yhql (approved):
tiny typo on "have" but it's alright :D

xchapron-ledger (commented):
Not sure about it, but I think you should bump the version of ledger-secure-sdk-sys used by ledger-device-sdk crate to use the last one?
To make sure that the sys crate doesn't get build twice with different versions?

cargo-ledger/src/main.rs
```diff
@@ -194,13 +194,31 @@ fn build_app(
...
+            args.push(String::from(
+                "--message-format=json-diagnostic-rendered-ansi",
+            ));
+
+            match std::env::var("LEDGER_SDK_PATH") {
+                Ok(_) => (),
+                Err(_) => match c_sdk_path {
+                    Ok(path) => args
+                        .push(format!("--config env.LEDGER_SDK_PATH={}", path)),
+                    Err(_) => println!("C SDK will habe to be cloned"),
```
agrojean-ledger, line 216: Typo `habe` -> `have`

agrojean-ledger (approved):

yogh333:
> Not sure about it, but I think you should bump the version of ledger-secure-sdk-sys used by ledger-device-sdk crate to use the last one?
> To make sure that the sys crate doesn't get build twice with differen

Yes you are right

PR #119, opened 2024-01-17
added new workflow to test updated nightly

yogh333:


.github/workflows/build_with_current_nightly.yml
```diff
@@ -0,0 +1,55 @@
...
+          mv ./target/${BUILD_DEVICE_NAME}/release/app-boilerplate-rust ./build/${BIN_DIR_NAME}/bin/app.elf
+      
+      - name: Upload binary artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: "app_elf_binaries"
+          path: app-boilerplate/build/*
+          if-no-files-found: error
+  
+  ragger_tests:
```
agrojean-ledger, line 48: You will need to run this in the dev-tools image (not the regular app-builder, it doesn't have Ragger nor Speculos)

yogh333: Shall be fixed in [reusable_ragger_tests](https://github.com/LedgerHQ/ledger-app-workflows/blob/master/.github/workflows/reusable_ragger_tests.yml) flow as it uses `ubuntu-latest` nowadays.

agrojean-ledger: Weird to use `ubuntu-latest` and then do 

```
      - name: Install tests dependencies
        run: |
          sudo apt-get update && sudo apt-get install -y qemu-user-static tesseract-ocr libtesseract-dev
          pip install -U pip setuptools
          pip install -r "${{ needs.call_get_app_metadata.outputs.pytest_directory }}/requirements.txt"
```

in the reusable workflow... Maybe it's because we want to be closer to a real host environment ? :thinking:

yogh333: Please see https://github.com/LedgerHQ/ledger-app-workflows/pull/56, I do not know why an ubuntu env is used to run test (rather than our dev-tools container)...

agrojean-ledger (approved):

PR #120, opened 2024-01-17
rework MultiFieldReview to use less stack

yhql:
Preallocating 48 Pages led to a huge stack usage and to crashes on Nano S. 
This gadget was reworked so that it does not preallocate, and makes it "simpler" to visualize Fields by giving them their own widget (similar to MessageScroller in behaviour, except with a counter and a different amount of lines used).

Note that `MAX_LINES` was changed to `4` (was previously `3`) for Nano X/S+ to avoid ifdefs to match the C version of the Field view, but as a consequence, a Menu now displays over 4 lines.

### DIFF
--- a/ledger_device_sdk/src/ui/gadgets.rs
+++ b/ledger_device_sdk/src/ui/gadgets.rs
@@ -1,7 +1,3 @@
-#![allow(dead_code)]
-
-use core::str::from_utf8;
-
 use crate::{
     buttons::ButtonEvent::*,
     io::{self, ApduHeader, Comm, Event, Reply},
@@ -92,10 +88,14 @@ pub fn display_pending_review(comm: &mut Comm) {
     clear_screen();
 
     // Add icon and text to match the C SDK equivalent.
-    if cfg!(target_os = "nanos") {
+    #[cfg(target_os = "nanos")]
+    {
         "Pending".place(Location::Custom(2), Layout::Centered, true);
         "Ledger review".place(Location::Custom(14), Layout::Centered, true);
-    } else {
+    }
+
+    #[cfg(not(target_os = "nanos"))]
+    {
         WARNING.draw(57, 10);
         "Pending".place(Location::Custom(28), Layout::Centered, true);
         "Ledger review".place(Location::Custom(42), Layout::Centered, true);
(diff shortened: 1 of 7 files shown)


ledger_device_sdk/src/ui/gadgets.rs
```diff
@@ -361,15 +365,16 @@ impl<'a> Menu<'a> {
...
-    BoldNormal,    // One line of bold text and one line of normal text.
-    Normal,        // 2 lines of centered text.
+    PictureBold, // Icon on top with one line of text on the bottom.
+    BoldNormal,  // One line of bold text and one line of normal text.
+    Normal,      // 2 lines of centered text.
 }
 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Default)]
 pub struct Page<'a> {
```
agrojean-ledger, line 378: As I understand `chunk_count` and `chunk_idx` are now useless fields in the `Page` struct as the `Field` manages the display of this information by itself. 

The logic in the `Page` `place` function that manages the display of chunk and chunk count for `PageStyle::BoldNormal` should be removed.

yogh333:
FYI I have tested this branch with app-boilerplate-rust (see [here](https://github.com/LedgerHQ/app-boilerplate-rust/tree/test/multifield_review_refactor)) and all tests are OK (with --golden_run option as some snapshots are modified due to [this](https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/120#discussion_r1455720035)). And good news, all tests still OK with the last nightly (17/01/2024).

ledger_device_sdk/src/ui/gadgets.rs
```diff
@@ -719,6 +724,98 @@ pub struct Field<'a> {
...
+            let mut chunks = [Label::default(); layout::MAX_LINES];
+            for (i, chunk) in self
+                .value
+                .as_bytes()
+                .chunks(MAX_CHAR_PER_LINE)
+                .skip(page * chunk_max_lines)
+                .take(chunk_max_lines)
+                .enumerate()
+            {
+                chunks[1 + i] = Label::from(core::str::from_utf8(chunk).unwrap());
```
dj8yfo, line 753: this will work for ascii strings, but may panic for arbitrary utf8

dj8yfo: current `"1.4.2"` `ledger_device_sdk` crashes on displaying a string in chinese? with `MultiFieldReview`:
```rust
"《東明聞見錄》.8.9.10"
```
and a Georgian one: 
```rust
გთხოვთ ახლავე გაიაროთ
```
,
but works with a sample string in russian and a German? one:

```rust
"„Anführungszeichen“ .8.9"  
```
displaying blanks instead of stranger characters

yhql: The SDK only supports ASCII characters anyway and would simply display blanks (or maybe crash) on anything else.
I'm not a fan of unwrapping, but rather than using `unsafe { from_utf8_unchecked() }` and justifying with a comment, I'd rather keep the code panicking on an incorrect string

yhql: But since it would not panic on unknown alphabets I replaced some of them with `unwrap_or("")`

ledger_device_sdk/src/ui/gadgets.rs
```diff
@@ -719,6 +724,98 @@ pub struct Field<'a> {
...
+
+            crate::ui::screen_util::screen_update();
+        };
+
+        draw(cur_page);
+
+        loop {
+            match get_event(&mut buttons) {
+                Some(ButtonEvent::LeftButtonPress) => {
+                    LEFT_S_ARROW.instant_display();
```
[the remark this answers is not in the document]
yhql: Guessing from @yogh333 's comment that this is okay in the end. I'd rather keep those

ledger_device_sdk/src/ui/gadgets.rs
yhql: Done

yogh333 (approved):
Bump version

PR #123, opened 2024-01-23
Feat/use nightly toolchain explicitly

yogh333:


yogh333:
Add a check in cargo-ledger to set the proper toolchain when building app with our `ledger-app-builder` container as the default toolchain is set to **stable** (this is required for some C apps using Rust to compile static libraries).

cargo-ledger/src/main.rs
```diff
@@ -201,6 +202,20 @@ fn build_app(
...
             let mut args: Vec<String> = vec![];
+            match std::env::var("RUST_NIGHTLY") {
+                Ok(version) => {
+                    println!("Use Rust nightly toolchain: {}", version);
+                    args.push(format!("+{}", version))
+                }
+                Err(_) => {
+                    let rustup_cmd =
+                        Command::new("rustup").arg("default").output().unwrap();
+                    print!(
```
xchapron-ledger, line 213: Shouldn't it be `println` like above?

ledger_secure_sdk_sys/build.rs
```diff
@@ -477,7 +477,7 @@ impl SDKBuilder {
         );
 
         let mut bindings = bindgen::builder()
-            .clang_args(&args)
```
xchapron-ledger, line 480: Do we need this change to be included in ledger-device-sdk crate?

xchapron-ledger: I'm afraid of apps using the last sys crate version in their cargo.toml as this end with building twice the crate with two different version, and generates an invalid app binary.

yogh333: ok let's postpone this modification for later then...

agrojean-ledger (approved):

PR #125, opened 2024-01-25
add workflow to build all Rust apps for every PR

yogh333:


.github/workflows/build_all_apps.yml
```diff
@@ -0,0 +1,86 @@
...
+jobs:    
+    test-build:
+        name: Build for all targets
+        strategy:
+            fail-fast: false
+            matrix:
+                include: 
+                    - repo: 'app-mobilecoin'
+                      branch: 'develop'
+                      build_path: './fw'
```
xchapron-ledger, line 22: Can we parse the `ledger_app.toml` manifest instead?

.github/workflows/build_all_apps.yml
```diff
@@ -0,0 +1,86 @@
...
+
+jobs:    
+    test-build:
+        name: Build for all targets
+        strategy:
+            fail-fast: false
+            matrix:
+                include: 
+                    - repo: 'app-mobilecoin'
+                      branch: 'develop'
```
xchapron-ledger, line 21: Can we use the default branch instead? Though this might be less clear...

.github/workflows/build_all_apps.yml
```diff
@@ -0,0 +1,86 @@
...
+                      build_path: './fw'
+                      device: nanosplus
```
xchapron-ledger, line 23: Could also use `ledger-app.toml` field?

yogh333: WIP

.github/workflows/build_all_apps.yml
yogh333: WIP

.github/workflows/build_all_apps.yml
```diff
@@ -0,0 +1,89 @@
...
+            fail-fast: false
+            matrix:
+                include: 
+                    #- repo: 'app-mobilecoin'
+                    #  branch: 'develop'
+                    - repo: 'app-radix-babylon'
+                      branch: 'develop'
+                    - repo: 'app-boilerplate-rust'
+                      branch: 'main'
+
```
tdejoigny-ledger, line 26: Could be good to not have an hardcoded list but to build dynamically the list. 
I tried to do that in this [PR](https://github.com/LedgerHQ/ledger-secure-sdk/pull/553/files) for C app (name: Get C apps repos)

tdejoigny-ledger: yes it could be a good idea

tdejoigny-ledger: ok

tdejoigny-ledger (approved):

.github/workflows/build_all_apps.yml
xchapron-ledger: If nobody add it to ledgered it's never going to be there ;)

PR #128, opened 2024-02-01
Jca/relocation obsidian way

yogh333:


cargo-ledger/src/setup.rs
```diff
@@ -13,7 +13,7 @@ pub fn install_targets() {
     let sysroot_cmd = std::str::from_utf8(&sysroot_cmd).unwrap().trim();
 
     let target_files_url = Path::new(
-        "https://raw.githubusercontent.com/LedgerHQ/ledger-device-rust-sdk/cee5644d6c20ff97b13e79a30caca751b7b52ac8/ledger_device_sdk/",
+        "https://raw.githubusercontent.com/LedgerHQ/ledger-device-rust-sdk/master/ledger_device_sdk/"
```
xchapron-ledger, line 16: I'm afraid that not targeting a specific version will create incompatibilities.

If at some point with edit nanos.json, it might end up incompatible with a previous version of cargo ledger, but if cargo ledger is configured to use the last version, then this version of cargo ledger will be unusable, breaking reproducibility.

yogh333: Yes fixed 👍

cargo-ledger/src/setup.rs
```diff
@@ -39,4 +39,42 @@ pub fn install_targets() {
             println!("* {target} already installed");
         }
     }
+
```
yhql, line 42: For those who already installed targets, `cargo ledger setup` won't be able to tell the difference between the old targets and the new ones and wont install the .json present in this patch. Should we handle that or just explain that the targets should be manually removed beforehand?

yogh333: Ok we should handle this yes. The most simple could be to overwrite systematically the targets even if they are existing ?

PR #130, opened 2024-02-08
Fix NVMData on S+; add a new symbol at start of .nvm_data

dfordivam:


yogh333:
I think that it should also be modified [there](https://github.com/LedgerHQ/ledger-device-rust-sdk/blob/6f6aa0a72317d004acee56f26fb0e42e8fb396b4/cargo-ledger/src/utils.rs#L63) also

PR #132, opened 2024-02-08
Wrap cryptolib functions

yogh333:
- Added wrapper for hash functions
- Return Chain Code (BIP32 derivation)


ledger_device_sdk/src/ecc.rs
```diff
@@ -377,7 +377,12 @@ impl<const P: usize> ECPublicKey<P, 'E'> {
 ///
 /// Checks consistency of curve choice and key length
 /// in order to prevent the underlying syscall from throwing
-pub fn bip32_derive(curve: CurvesId, path: &[u32], key: &mut [u8]) -> Result<(), CxError> {
+pub fn bip32_derive(
+    curve: CurvesId,
+    path: &[u32],
+    key: &mut [u8],
+    cc: &mut [u8],
+) -> Result<(), CxError> {
```
xchapron-ledger, line 385: I don't get why for ed25519 the requested length is 96 few lines below?

ledger_device_sdk/src/ecc.rs
```diff
@@ -377,7 +377,12 @@ impl<const P: usize> ECPublicKey<P, 'E'> {
 ///
 /// Checks consistency of curve choice and key length
 /// in order to prevent the underlying syscall from throwing
```
xchapron-ledger, line 379: Note that the syscall can throw if the requested curve or path haven't been declared in the manifest.

ledger_device_sdk/src/ecc/stark.rs
```diff
@@ -16,8 +16,9 @@ const C_CX_STARK256_N: [u8; 32] = [
 /// https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2645.md
 pub fn eip2645_derive(path: &[u32], key: &mut [u8]) {
     let mut x_key = Secret::<64>::new();
+    let mut cc = Secret::<32>::new();
     // Ignoring 'Result' here because known to be valid
-    let _ = super::bip32_derive(CurvesId::Secp256k1, path, x_key.as_mut());
+    let _ = super::bip32_derive(CurvesId::Secp256k1, path, x_key.as_mut(), cc.as_mut());
```
xchapron-ledger, line 21: if cc is not used below, maybe you can avoid allocating it for nothing?

ledger_device_sdk/src/ecc.rs
yogh333: Agreed, looks to be ok with 64 🤔. Maybe a typo @yhql ?

ledger_device_sdk/src/ecc.rs
yogh333: `no_throw` version is not available through bindings as it is defined as a static inline function in C SDK.

xchapron-ledger: My remark was just about the fact that the comment isn't perfectly true... There is no way to avoid the syscall to throw.

ledger_device_sdk/src/ecc.rs
yogh333: Fix in [this commit](https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/132/commits/5b63e3e971e242e5a43304f4459c467374d23c67), waiting feedback from @yhql if revert is needed

ledger_device_sdk/src/testing.rs
```diff
@@ -53,28 +53,26 @@ pub struct TestType {
...
 #[cfg(feature = "speculos")]
 pub fn sdk_test_runner(tests: &[&TestType]) {
-    use core::ffi::c_void;
-    use ledger_secure_sdk_sys::{pic, pic_rs};
     let mut failures = 0;
     debug_print("--- Tests ---\n");
     for test_ in tests {
         // (ノಠ益ಠ)ノ彡ꓛIꓒ
-        let test = pic_rs(*test_);
+        let test = *test_;
```
niooss-ledger, line 60: Why did this Pull Request remove the `PIC` macro? It feels like such a change should be split in a different PR and reviewed on its own (because I am not knowledgeable enough about the state of application loading and whether we still need `PIC` (and `pic_rs`) in embedded applications.

yogh333: Well, `PIC` macro are no more needed following #128, so I did a little cleanup in the test runner used on `speculos`. I agree that it could have been proposed as a different PR but the impact is quite limited as it only concerns tests on speculos.

niooss-ledger (approved):
Looks good to me.

yhql (approved):
LGTM, would recommend using `macro_rules!` to declare `HashInit` for all variants though

PR #133, opened 2024-02-09
Bump stack size

siy:
Use RAM for stack. 
For Nano S - as much as possible
For Nano X/S Plus - 16K (some RAM is still unused, but should be OK for now)


ledger_device_sdk/nanos_layout.ld
```diff
@@ -6,5 +6,5 @@ MEMORY
 }
 
 PAGE_SIZE  = 64;
-STACK_SIZE = DEFINED(stack_size) ? stack_size : 1024;
+STACK_SIZE = 3692+512;
```
xchapron-ledger, line 9: This is app specific. It can be set like this in the SDK.
Current linker will put as much stack as possible (filling the full RAM), with a minimum of STACK_SIZE.

siy: For some reason, an application built with default link files crashes on Nano S. the Same application built with modified linker file works just fine. [Here](https://github.com/radixdlt/babylon-ledger-app/actions/runs/7871211167/job/21474101644?pr=62#step:8:268) is an example of such a crash in speculos.

PR #135, opened 2024-02-22
Allow cloning ECPublicKey and make its fields pub

dfordivam:
The contents of the public key are already pub, so it should be ok to clone it.


yogh333 (commented):
LGTM

yogh333:
Also could you please bump version 1.5.2 for `ledger_device_sdk` ?

dfordivam:
@yogh333 as part of this PR? (I suppose it should be done separately no?)

yogh333 (changes requested):
@dfordivam yes bump of new version will be made later. In the meantime, could you please rebase your branch 🙏

yogh333 (approved):

Issue #136, opened 2024-02-22
Add support for Ed25519 SLIP10 derivation


dfordivam:
Here is an example implementation
Ref: https://github.com/satoshilabs/slips/blob/master/slip-0010.md

```
fn ed25519_derive_from_path_slip10(path: &[u32]) -> ECPrivateKey<32, 'E'> {
    let mut tmp = [0; 96];
    let seed_key: &mut [u8; 12] = &mut [0; 12];
    seed_key.copy_from_slice(b"ed25519 seed");
    unsafe {
        os_perso_derive_node_with_seed_key(
            HDW_ED25519_SLIP10,
            CurvesId::Ed25519 as u8,
            path.as_ptr(),
            path.len() as u32,
            tmp.as_mut_ptr(),
            core::ptr::null_mut(),
            seed_key.as_mut_ptr(),
            12,
        );
    }
    let mut sk = ECPrivateKey::new(CurvesId::Ed25519);
    let keylen = sk.key.len();
    sk.key.copy_from_slice(&tmp[..keylen]);
    tmp.copy_from_slice(&[0; 96]);
    sk
}
```

PR #139, opened 2024-02-28
Add SLIP10 derivation support for Ed25519

yogh333:
See #136

### DIFF
--- a/ledger_device_sdk/src/ecc.rs
+++ b/ledger_device_sdk/src/ecc.rs
@@ -523,6 +523,31 @@ impl SeedDerive for Ed25519 {
     }
 }
 
+/// Support SLIP10 derivation for Ed25519
+impl Ed25519 {
+    pub fn derive_from_path_slip10(path: &[u32]) -> ECPrivateKey<32, 'E'> {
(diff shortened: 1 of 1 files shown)


ledger_device_sdk/src/ecc.rs
```diff
@@ -523,6 +523,31 @@ impl SeedDerive for Ed25519 {
     }
 }
 
+/// Support SLIP10 derivation for Ed25519
+impl Ed25519 {
+    pub fn derive_from_path_slip10(path: &[u32]) -> ECPrivateKey<32, 'E'> {
+        let mut tmp = Secret::<64>::new();
+        let seed_key: &mut [u8; 12] = &mut [0; 12];
+        seed_key.copy_from_slice(b"ed25519 seed");
```
xchapron-ledger, line 531: This is the default value in the Os, so doesn't needed to be pass to os_perso_derive_node_with_seed_key

xchapron-ledger (commented):
I'm not sure, but I think that ledgerctl doesn't support slip10 path install parameter ?

yhql (approved):

PR #142, opened 2024-03-05
Add bool param to cargo ledger to allow generating the standard build…

agrojean-ledger:
Add bool param to cargo ledger to allow generating the standard build tree (used in C SDK as well)

### DIFF
--- a/cargo-ledger/src/main.rs
+++ b/cargo-ledger/src/main.rs
@@ -45,7 +45,7 @@ struct DeviceMetadata {
 #[command(name = "cargo")]
 #[command(bin_name = "cargo")]
 #[clap(name = "Ledger devices build and load commands")]
-#[clap(version = "0.0")]
+#[clap(version = "1.4.0")]
 #[clap(about = "Builds the project and emits a JSON manifest for ledgerctl.")]
 enum Cli {
     Ledger(CliArgs),
@@ -92,6 +92,16 @@ impl AsRef<str> for Device {
     }
 }
 
+impl Device {
+    fn as_ref_for_std_tree(&self) -> &str {
+        match self {
+            Device::Nanos => "nanos",
+            Device::Nanox => "nanox",
+            Device::Nanosplus => "nanos2",
+        }
+    }
+}
+
 #[derive(Subcommand, Debug)]
 enum MainCommand {
     #[clap(about = "install custom target files")]
@@ -104,6 +114,11 @@ enum MainCommand {
         #[clap(short, long)]
         #[clap(help = "load on a device")]
         load: bool,
+        #[clap(short, long)]
+        #[clap(
+            help = "copy output files (.hex, .elf, .apdu) to standard ledger app build tree (inspired from C SDK)"
+        )]
+        standard_build_tree: bool,
         #[clap(last = true)]
         remaining_args: Vec<String>,
     },
@@ -117,9 +132,10 @@ fn main() {
         MainCommand::Build {
             device: d,
             load: a,
+            standard_build_tree: s,
             remaining_args: r,
         } => {
-            build_app(d, a, cli.use_prebuilt, cli.hex_next_to_json, r);
+            build_app(d, a, s, cli.use_prebuilt, cli.hex_next_to_json, r);
         }
     }
 }
@@ -191,6 +207,7 @@ fn build_app(
     is_load: bool,
     use_prebuilt: Option<PathBuf>,
     hex_next_to_json: bool,
+    standard_build_tree: bool,
     remaining_args: Vec<String>,
 ) {
     let exe_path = match use_prebuilt {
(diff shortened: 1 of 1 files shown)


cargo-ledger/src/main.rs
```diff
@@ -117,9 +132,10 @@ fn main() {
         MainCommand::Build {
             device: d,
             load: a,
+            standard_build_tree: s,
             remaining_args: r,
         } => {
-            build_app(d, a, cli.use_prebuilt, cli.hex_next_to_json, r);
+            build_app(d, a, s, cli.use_prebuilt, cli.hex_next_to_json, r);
```
yogh333, line 138: Are `use_prebuilt` and `hex_next_to_json`flags really needed 🤔 ?

cargo-ledger/src/main.rs
```diff
@@ -363,6 +380,46 @@ fn build_app(
...
+                    err
+                );
+            }
+            copy_file_with_err_msg(&exe_path, &elf_dest);
+            copy_file_with_err_msg(&hex_file_abs, &hex_dest);
+            copy_file_with_err_msg(&apdu_file_path, &apdu_dest);
+        };
+        // Copying binaries to standard bin directory.
+        let bin_dir = PathBuf::new().join(current_dir).join("bin");
+        copy_binaries_to_standard_path(&bin_dir);
```
yogh333, line 413: What is this directory ? Shall we also copy all binaries (`hex`, `apdu` and `elf`) into a `/bin` directory ?

cargo-ledger/src/main.rs
```diff
@@ -363,6 +380,46 @@ fn build_app(
...
+        // Copying binaries to standard device build directory.
+        let target_build_dir = PathBuf::new()
+            .join(current_dir)
+            .join("build")
+            .join(device.as_ref_for_std_tree())
+            .join("bin");
+        copy_binaries_to_standard_path(&target_build_dir);
```
yogh333, line 420: Copy all binaries into a C-SDK like build tree, so do not understand about the `/bin` folder above...

cargo-ledger/src/main.rs
xchapron-ledger: In c app, artifcat are both under bin/ and build/nanox/bin/

cargo-ledger/src/main.rs
agrojean-ledger: We have both in the C SDK actually, the `bin` directory with the binaries from the latest `make` (files are overwritten every time, whatever the target selected) and the `build/<target>/bin/` tree that keeps the files of every target we built the app for.

cargo-ledger/src/main.rs
agrojean-ledger: I need to check where we actually use those, I think it's in the deployment CI, maybe we can get rid of them.

cargo-ledger/src/main.rs
agrojean-ledger: Some of our tools might use the `bin` and others the `build/<target>/bin`

yogh333: So for Rust, I am afraid that it is quite redundant and confusing. Having three (!) different folders with binaries, it is a lot...I was understanding that the `build` folder is required for other tools (ragger for instance). In Rust, both C SDK folders `bin` and `build` folders are merged into `target` folder as it already contains all the binaries for the different targets and it is updated at every build for the current target. So I do not see any reason to have the `bin` folder.

yogh333: Could we list the tools that need either `bin` or `build` folders ?

agrojean-ledger: It will never be an issue as it is opt-in only (enabled by the flag, optionally). What's more it will only ever be used by : 

* Our team internally (we know what we do),
* External developers who know what they do.

If it is used by people who don't know what they do, yes it might get confusing, but I don't really mind.

Also, we should stop thinking about C apps vs Rust apps. We have apps, and they should all integrate with our tooling in the same way. I personally don't want to lose time looking over every piece of code that our apps integrate with and justify our team choices over the past years that led to having `bin`  and `build`. It is the state of things today, and all apps should be able to provide this.

@yogh333 if you want to spend time doing the archaeology, go ahead. If you want to change how things are done, please do so as well, but do it for all tools and apps (so please update the C SDK as well).

cargo-ledger/src/main.rs
yhql: This is used (and contributed to) by Obsidian Systems

cargo-ledger/src/main.rs
yogh333: Hi, 

 I just would like to understand why we have to bend cargo-ledger to match with some obscure C-SDK constraints. I like understanding as, in general, it helps to take good and future-proof decision. I do not think it is a waste of time.
And yes we could change how things are done if it is better, but for this we need to **understand**.
I am still ok with the opt flag, as we agreed collectively.
Last but not least, I think that we still still a `cargo ledger clean` if our `cargo ledger build` creates some aditionnal non native Rust build folders.

agrojean-ledger: Agree about `cargo ledger clean`

PR #143, opened 2024-03-15
EdDSA signature creation via streaming

### LINKED ISSUE #137: Add EdDSA signature creation via streaming for arbitrary size messages
Certain blockchains (like Pocket) require signature of the entire message payload ( instead of the hash of the message). Doing such a signature generation for an arbitrary size message is not possible on the Ledger devices due to limited stack space. 

The only way to create signatures in limited stack space is by incrementally computing the SHA512 of the message.
Ref: https://datatracker.ietf.org/doc/html/rfc8032#page-13

The Ledger SDK must provide a way to compute EdDSA signature by consuming the message data incrementally, similar to the `Hash` interface.

A reference implementation is here https://github.com/alamgu/ledger-crypto-helpers/blob/8f46275fc7fe9b8169b4f689ce2a203fb02c993d/src/ed25519.rs#L72C6-L72C13

dfordivam:
Fixes #137

### DIFF
--- a/ledger_device_sdk/src/ecc.rs
+++ b/ledger_device_sdk/src/ecc.rs
@@ -1,3 +1,5 @@
+use crate::hash::sha2::Sha2_512;
+use crate::hash::{HashError, HashInit};
 use ledger_secure_sdk_sys::*;
 use zeroize::Zeroize;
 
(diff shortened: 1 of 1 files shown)


ledger_device_sdk/src/hash.rs
```diff
@@ -67,7 +67,7 @@ pub trait HashInit: Sized {
             Ok(())
         }
     }
-    fn finalize(mut self, output: &mut [u8]) -> Result<(), HashError> {
+    fn finalize(&mut self, output: &mut [u8]) -> Result<(), HashError> {
         let output_size = self.get_size();
```
yogh333, line 71: I do not think this is required:when a hash is finalized, then its context shall be consumed. To compute a new hash, then `HashInit::new()` shall be called. It is the same behaviour as with `HashInit::hash()`

ledger_device_sdk/src/hash.rs
```diff
@@ -67,7 +67,7 @@ pub trait HashInit: Sized {
             Ok(())
         }
     }
-    fn finalize(mut self, output: &mut [u8]) -> Result<(), HashError> {
+    fn finalize(&mut self, output: &mut [u8]) -> Result<(), HashError> {
```
yogh333, line 70: I do not think this is required:when a hash is finalized, then its context shall be consumed. To compute a new hash, then HashInit::new() shall be called. It is the same behaviour as with HashInit::hash()

yogh333:
Hi, please check #144 as a new proposal

ledger_device_sdk/src/hash.rs
dfordivam: Hi, I have removed this commit, now the PR uses `core::mem::take`

yogh333:
Closed by #144

PR #144, opened 2024-03-18
Yogh333/support ed25519 pure ed dsa

yogh333:
From #143 proposal

### DIFF
--- a/ledger_device_sdk/src/ecc.rs
+++ b/ledger_device_sdk/src/ecc.rs
@@ -1,6 +1,8 @@
 use ledger_secure_sdk_sys::*;
 use zeroize::Zeroize;
 
+use crate::hash::{sha2::Sha2_512, HashInit};
+
 mod stark;
 
 #[repr(u8)]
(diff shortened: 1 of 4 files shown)


dfordivam (commented):
There are a couple of problems with this approach, first the stack usage will be higher due to allocation of temporary `hash`, as compared to using the `self.hash`. The size of `Sha2_512` is 204 bytes. If only `self.hash` is used then stack usage could be 2x that, ie 408, due to construction of new() hash when doing a `take`. But if additional `hash` is allocated on stack in the `stream_sign` API, this could be 4x.

The second issue is that it is rather difficult to specify the `is_last` when doing streaming. It will require more complex logic for the app to figure this out, as compared to the use of a `finalize` step.

dfordivam:
Doing `stream_sign` with empty `msg` and `is_last = true` should work similar to the `finalize` step right? in that case this design is ok. It would be nice to have a test to confirm this

yogh333:
@dfordivam I have removed the `is_last` flag. Regarding the stack usage, I am not sure I fully understand your point: in this proposal, the stack footprint due to hash ctx is no more than 2x204 bytes = 408.

dfordivam:
@yogh333 I meant to say that the `ctx` already contains the hash, and an additional hash is allocated on the stack for temporary computations. The rust already allocates 2x space on stack for a single usage of hash, (due to the new/drop symantics). Therefore I suspect that this implementation would consume more stack (possibly 4x size of hash), vs reusing the hash from `ctx` (which should ideally consume 2x size of hash)
You will have to do stack size analysis to get the real numbers, as it is hard to predict the numbers.

dfordivam:
Hi @yogh333 
I tried your PR with the pocket app, and it works on S+, but S fails due to stack size overflow.
I have compared stack usage of this PR with mine (#143 )
143 -> Maximum stack: Some(3584)
144 -> Maximum stack: Some(3712)
So there is more room for stack size improvements.

yogh333:
@dfordivam  My last 2 commits should slightly improve stack usage 🙏

dfordivam:
@yogh333 the PR now works with pocket nanos!
Maximum stack: Some(3600)

yogh333:
@dfordivam A last modification that enable to retrieve the whole (R,S) signature in a single buffer without modifying usage stack. Please approve the PR if it is ok.
@yhql Could you please review and approve the PR 🙏

ledger_device_sdk/src/ecc.rs
```diff
@@ -283,9 +285,234 @@ impl<const N: usize> ECPrivateKey<N, 'W'> {
...
+macro_rules! check_cx_ok {
+    ($fn_call:expr) => {{
+        let err = unsafe { $fn_call };
+        if err != CX_OK {
+            return Err(err.into());
+        }
+    }};
+}
+
+impl<const N: usize> ECPrivateKey<N, 'E'> {
```
yhql, line 315: `N` is actually fixed in this case, right? It could be removed

ledger_device_sdk/src/ecc.rs
```diff
@@ -283,9 +285,234 @@ impl<const N: usize> ECPrivateKey<N, 'W'> {
...
+    pub fn stream_sign_init(&self, ctx: &mut Ed25519Stream) -> Result<(), CxError> {
+        // Compute prefix (see https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.6, step 1)
+        match ctx.big_r_started {
```
yhql, line 318: Perhaps having two different types, one for the initialized state and the other one that can be finalized, would help here.
Should `stream_sign_init` really be part of ECPrivateKey instead of Ed25519Stream? 
```
impl Ed25519StreamInit {
    fn stream_sign_init(self, key: &ECPrivateKey<64, 'E'>) -> Ed25519Stream {}
 }
 
 impl Ed25519Stream {
     fn sign_finalize()
     fn update()
 }
 ```

Comment (yhql) [reply ledger_device_sdk/src/ecc.rs:318]: Proposal for this https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/147

**Review by yhql — APPROVED** [review]

**Review by dfordivam — CHANGES_REQUESTED** [review]
Hi @yogh333 
Unfortunately with the latest changes the stack size usage has increased a lot.
It could be due to a number of reasons.
Maximum stack: Some(4568)

**Comment by yhql** [comment]
> Hi @yogh333 Unfortunately with the latest changes the stack size usage has increased a lot. It could be due to a number of reasons. Maximum stack: Some(4568)

Sorry about this, I was not expecting such a blowup. https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/148 saves about 800 bytes of stack usage on an example taken from the corresponding test (700-something on `::new()` and 100 on `finalize`)

**Comment by dfordivam** [comment]
Hi, there has been some improvement with recent changes, but for me the number is still quite high.
Maximum stack: Some(4064)

Also I suggest changing the `new` API to an `init` API, as we generally need to pre-allocate the `Ed25519Stream` on stack.
```
pub fn init(&mut self, key: &ECPrivateKey<32, 'E'>) -> Result<(), CxError>
```
With this change I managed to get slight improvements.
Maximum stack: Some(4008)

**Comment by yogh333** [comment]
@dfordivam Regarding your last proposal, is it sufficient to be able to run the Pocket app on Nano S ?
And could you make a last comparison with this [commit](https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/144/commits/ac829a8ac69867989a9ae4fcdbde5e95a3521e0b) which was ok for Nano S while having the lowest stack usage (3600). Thank you !

**Comment by yogh333** [comment]
@dfordivam Please make a test with these last changes 🙏 and tell if it is sufficient for Pocket on Nano S. If not, I am afraid we shall revert to this [commit](https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/144/commits/ac829a8ac69867989a9ae4fcdbde5e95a3521e0b).
Thank you !

**Comment by dfordivam** [comment]
@yogh333 
I debugged a bit and figured out that the following change reduces max stack size from 4008 to 3585.
The biggest stack size was `core::ptr::drop_in_place`

With this change the pocket app works on nanos

```
diff --git a/ledger_device_sdk/src/ecc.rs b/ledger_device_sdk/src/ecc.rs
index 72409a0..7730ab1 100644
--- a/ledger_device_sdk/src/ecc.rs
+++ b/ledger_device_sdk/src/ecc.rs
@@ -294,7 +294,7 @@ pub struct Ed25519Stream {
 impl Default for Ed25519Stream {
     fn default() -> Self {
         Ed25519Stream {
-            hash: Sha2_512::new(),
+            hash: Sha2_512::default(),
             big_r: [0u8; 32],
             signature: [0u8; 64],
         }
@@ -314,6 +314,7 @@ impl Ed25519Stream {
     pub fn init(&mut self, key: &ECPrivateKey<32, 'E'>) -> Result<(), CxError> {
         // Compute prefix (see https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.6, step 1)
         let mut temp = Secret::<64>::new();
+        self.hash.reset();
         self.hash
             .hash(&key.key[..], temp.as_mut())
             .map_err(|_| CxError::GenericError)?;
```

PR #147, opened 2024-03-21
Proposal for eddsa streamer API

yhql:
- Remove internal tracking of initialized/uninitialized state by having two separate structures
- Invert logic by having the `Ed25519Stream` implement the stream_sign functions, not the private key itself


yogh333 (approved):
LGTM

PR #148, opened 2024-03-22
Save stack during eddsa streaming

yhql:
- Remove useless StreamInit struct and provide a `new()` to Stream instead
- Use the ed25519's default hash as temporary during new/init
- Avoid making Hash copyable, fix a method that takes `mut self` instead of `&mut self` and induced copies

### DIFF
--- a/ledger_device_sdk/src/hash.rs
+++ b/ledger_device_sdk/src/hash.rs
@@ -67,7 +67,7 @@ pub trait HashInit: Sized {
             Ok(())
         }
     }
-    fn finalize(mut self, output: &mut [u8]) -> Result<(), HashError> {
+    fn finalize(&mut self, output: &mut [u8]) -> Result<(), HashError> {
         let output_size = self.get_size();
         if output_size > output.len() {
             return Err(HashError::InvalidOutputLength);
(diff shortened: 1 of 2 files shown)


ledger_device_sdk/src/hash.rs
```diff
@@ -37,7 +37,7 @@ pub trait HashInit: Sized {
     fn get_size(&mut self) -> usize {
         unsafe { cx_hash_get_size(self.as_ctx()) }
     }
-    fn hash(mut self, input: &[u8], output: &mut [u8]) -> Result<(), HashError> {
+    fn hash(&mut self, input: &[u8], output: &mut [u8]) -> Result<(), HashError> {
```
yogh333, line 40: When calling `hash()`, the context `self` should be consumed to prevent any further usage later by the caller (same as `finalize()`): that why it is passed as a value and not as a ref The only use case where the context is reused is `update()`. I agree that it prevents a copy to be performed however...

yogh333: To be consistent, if we keep this change, it shall be also performed for `finalize()`

yogh333: Might be solved by https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/143#discussion_r1529664279 ?

ledger_device_sdk/src/hash.rs
```diff
@@ -84,7 +84,7 @@ pub trait HashInit: Sized {
 
 macro_rules! impl_hash {
     ($typename:ident, $ctxname:ident, $initfname:ident, $size:expr) => {
-        #[derive(Copy, Clone, Default)]
+        #[derive(Default)]
```
yogh333, line 87: Same below ?

PR #150, opened 2024-03-26
feat: make `MultiFieldReview::show()` return error on non-ASCII input strings

### LINKED ISSUE #124: Suggestion: declare explicit support of `ASCII`-only charset at SDK level
this line https://github.com/LedgerHQ/ledger-device-rust-sdk/blob/master/ledger_device_sdk/src/ui/gadgets.rs#L831 can be changed to 
```rust
enum FieldError {
    NonASCIIChar, 
}

pub fn show(&self) -> Result<bool, FieldError> {
...
```

Probably two additional checks can be added after line https://github.com/LedgerHQ/ledger-device-rust-sdk/blob/master/ledger_device_sdk/src/ui/gadgets.rs#L693 to assert that `name` and `value` fields of `Field` are comprised only of ASCII chars (https://doc.rust-lang.org/std/primitive.str.html#method.is_ascii), and returning `Err(FieldError::NonASCIIChar)` otherwise.

It appears to me that reflecting the fact, that non-ascii charsets are not supported, with types is more robust than displaying randomly blank chunks without any indication to user, why this happens.

`MultiFieldReview::show()` can be unwrapped as well at the call sites, if `ledger-device-rust-sdk` user doesn't want to bother with handling errors, 
or the known error variants can be matched and handled, and e.g. a widget displayed to user, stating that `Error: non-ASCII payload is not supported` and a corresponding APDU status returned, refusing to continue to process.

dj8yfo:
resolves #124 
dependent pr : https://github.com/dj8yfo/app-near-rs/pull/49
This variant is the most ergonomic one in terms of code changes required client-side among 3 variants attempted

### DIFF
--- a/ledger_device_sdk/src/ui/gadgets.rs
+++ b/ledger_device_sdk/src/ui/gadgets.rs
@@ -690,7 +690,11 @@ pub struct Field<'a> {
 }
 
 impl<'a> Field<'a> {
-    pub fn event_loop(&self, incoming_direction: ButtonEvent) -> ButtonEvent {
+    /// # Panics
+    ///
+    /// This method may panic if either `self.name` or `self.value` contains some non-ASCII
+    /// characters. 
+    fn event_loop(&self, incoming_direction: ButtonEvent) -> ButtonEvent {
         let mut buttons = ButtonsState::new();
         let chunk_max_lines = layout::MAX_LINES - 1;
         let page_count = 1 + self.value.len() / (chunk_max_lines * MAX_CHAR_PER_LINE);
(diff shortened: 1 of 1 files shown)


ledger_device_sdk/src/ui/gadgets.rs
```diff
@@ -714,7 +718,7 @@ impl<'a> Field<'a> {
                 .take(chunk_max_lines)
                 .enumerate()
             {
-                chunks[1 + i] = Label::from(core::str::from_utf8(chunk).unwrap_or(""));
+                chunks[1 + i] = Label::from(core::str::from_utf8(chunk).unwrap());
```
dj8yfo, line 721: is assumption about ASCII-only `self.name` and `self.value` holds true, then `core::str::from_utf8(...)` isn't an `Err` for any slice of `self.name` and `self.value`

PR #153, opened 2024-04-10
Jca/fix syscall change api level 18

yogh333:


xchapron-ledger (commented):
Is it retrocompatible with previous version? Cause we need to be able to build with previous version too.

yogh333:
Yes I can build for API_LEVEL_5 as well

PR #154, opened 2024-05-03
remove invalid workspace-wide configuration

yhql:
This top-level configuration applies to all crates in the workspace, but it was invalid for the ones that are not meant to compile as an embedded app. (see https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure "At present, when being invoked from a workspace, Cargo does not read config files from crates within the workspace")

`cargo install --path cargo-ledger` works again from the workspace with this fix


kingofpayne:
I confirm installing cargo-ledger works by removing `.cargo/config.toml`

kingofpayne (approved):

yogh333 (approved):

PR #155, opened 2024-05-21
Heap and global allocator

kingofpayne:
Experimental heap and global allocator which can be optionaly enabled with "heap" feature.
This allows using for instance: String, format!, Vec... Current heap size is 8K.


ledger_device_sdk/src/lib.rs
```diff
@@ -52,6 +53,43 @@ macro_rules! set_panic {
...
+    unsafe fn release(restore_state: RawRestoreState) {}
+}
+
+/// Initializes the heap memory for the global allocator.
+///
+/// The heap is stored in the stack, and has a fixed size.
+/// This method is called just before [sample_main].
+#[no_mangle]
+#[cfg(feature = "heap")]
+extern "C" fn heap_init() {
```
yogh333, line 83: Question: as it looks like `heap_init()`is only used by the sys crate, why not defining this function in the sys crate as well ?

ledger_device_sdk/src/lib.rs
```diff
@@ -52,6 +53,43 @@ macro_rules! set_panic {
...
+    const HEAP_SIZE: usize = 8192;
```
yogh333, line 84: Could we make this size configurable ?

ledger_device_sdk/src/lib.rs
kingofpayne: Good suggestion. I moved all the heap code to `ledger_secure_sdk_sys`.

ledger_device_sdk/src/lib.rs
kingofpayne: Indeed it would be better if this was configurable by the SDK users.

For the moment I am not sure how to make it configurable, I currently know two possibilities:
- pass the heap size as an environment variable during the compilation (defined in `.cargo/config.toml`), fetched in the SDK code using rust `env!` macro,
- use crate features and have predefined possible heap sizes (`heap-1k`, `heap-2k`, `heap-4k`, `heap-8k`, etc.).

First solution offers more flexibility, but seems to be a bit hacky. I find second option cleaner, but it restricts the possible heap sizes.

For the moment I suggest keeping a fixed value that should work for 90% of the cases, and see later how to improve that. If heap size is too big, users can still disable the feature and work without heap as of today.

yogh333 (approved):

ledger_secure_sdk_sys/src/lib.rs
```diff
@@ -33,4 +35,41 @@ pub fn pic_rs_mut<T>(x: &mut T) -> &mut T {
...
+}
+
+/// Initializes the heap memory for the global allocator.
+///
+/// The heap is stored in the stack, and has a fixed size.
+/// This method is called just before [sample_main].
+#[no_mangle]
+#[cfg(feature = "heap")]
+extern "C" fn heap_init() {
+    const HEAP_SIZE: usize = 8192;
```
agrojean-ledger, line 66: Why not use an env variable with a default value defined in the Cargo.toml of the SDK : https://doc.rust-lang.org/cargo/reference/config.html#env ?

It could be overwritten at compile time by an "advanced" user : `HEAP_SIZE=4096 cargo ledger build stax` and it would be parsed in the the lib (I know `env!` outputs a `str` but maybe there's a workaround we can implement) 

@yogh333 @kingofpayne

kingofpayne: Yes this is probably the best way to go. It requires converting the str to an int as a const fn, which is a bit hacky but nonetheless possible.
Also the `HEAP_SIZE` environment variable should better be defined in the `.cargo/config.toml` file.

yogh333:
I did some tests with the Rust boilerplate (see https://github.com/LedgerHQ/app-boilerplate-rust/pull/58) and with `HEAP_SIZE=8K` `Ragger` tests are ok with S+, X, Stax and Flex. However, for Nano S, it does not even compile, the relocation script fails (5000 bytes overflow). I have tried naively to decrease the heap size down to `2048` bytes, compilation is ok for Nano S but I have core dumped during Ragger tests. With '1024` bytes it works, but not sure it will be enough for any other app (Obsidian, Radix for instance)...

ledger_secure_sdk_sys/src/lib.rs
agrojean-ledger: @kingofpayne Is it possible for you to test this solution before we merge the PR ? It would be nice to have this option from the get go.

agrojean-ledger: There's a solution I've found on stackoverflow (https://stackoverflow.com/questions/37526598/how-can-i-override-a-constant-via-a-compiler-option) for the conversion from str to int but I'm not sure this is how we want to do it as it requires another dependency on the `lazy_static` crate :

```
const MAX_DIMENSIONS_RAW: &'static str = env!("MAX_DIMENSIONS");
lazy_static! {
    static ref MAX_DIMENSIONS: usize = MAX_DIMENSIONS_RAW.parse().unwrap();
}
```

agrojean-ledger: Let's deal with this in a separate PR. I created an issue in our internal Rust SDK Jira board : https://ledgerhq.atlassian.net/browse/EMBSDK-50

agrojean-ledger (approved):

PR #157, opened 2024-05-28
Cleanup and simplify NBGL structs

yogh333:


ledger_device_sdk/src/nbgl.rs
```diff
@@ -177,32 +177,30 @@ const INFO_FIELDS: [*const c_char; 2] = [
 /// Used to display the home screen of the application, with an optional glyph,
 /// information fields, and settings switches.  
 pub struct NbglHomeAndSettings<'a> {
-    app_name: *const c_char,
-    icon: *const nbgl_icon_details_t,
```
agrojean-ledger, line 181: Not sure replacing the icon field with a local variable in the `show` function will always work (both in `debug` and `release` compilation). I had many issues when passing local variables to the NBGL API where it made the app crash or had display issues. 

Did you verify the behavior of the app in functional tests (preferrably with the `release`  mode binary) with this change ?

ledger_device_sdk/src/nbgl.rs
```diff
@@ -305,16 +288,23 @@ impl<'a> NbglHomeAndSettings<'a> {
             }
 
             loop {
-                if self.glyph.is_some() {
-                    self.icon = &self.glyph.unwrap().into() as *const nbgl_icon_details_t;
-                }
+                let info_list: nbgl_contentInfoList_t = nbgl_contentInfoList_t {
```
agrojean-ledger, line 291: Same remark as with the icon, I had issues when passing a local content variable to the NBGL API. Please verify the behavior of the app in functional tests with this change.

ledger_device_sdk/src/nbgl.rs
```diff
@@ -492,26 +482,37 @@ impl<'a> NbglAddressReview<'a> {
...
-            let c_verif_str = c_string_helper.to_cstring(self.verify_str).unwrap();
-
-            if self.glyph.is_some() {
-                self.icon = &self.glyph.unwrap().into() as *const nbgl_icon_details_t;
-            }
+            //let c_string_helper = CStringHelper::<128>::new();
+            //let c_addr_str = c_string_helper.to_cstring(address).unwrap();
+            //let c_verif_str = c_string_helper.to_cstring(self.verify_str).unwrap();
+
+            let addr = CStr::from_bytes_until_nul(address.as_bytes())
```
agrojean-ledger, line 500: My opinion is we keep the C string helper everywhere in NBGL until we fix it with the global allocator. 

Also, I think string manipulations to add a `\0` should always be transparent to the user (that's why we have the string helper in the first place, something we had extensively discussed with @kingofpayne).

ledger_device_sdk/src/nbgl.rs
```diff
@@ -492,26 +482,37 @@ impl<'a> NbglAddressReview<'a> {
...
+                .unwrap()
+                .as_ptr();
+
+            let icon = if self.glyph.is_some() {
```
agrojean-ledger, line 504: I think I tried this syntax (declaring a local variable, which is more logical in this case) but had display issues or crashes.

ledger_device_sdk/src/ui/gadgets.rs
```diff
@@ -568,7 +578,17 @@ impl<'a> MultiPageMenu<'a> {
                     }
                 },
                 io::Event::Command(ins) => return EventOrPageIndex::Event(io::Event::Command(ins)),
-                _ => (),
+                io::Event::Ticker => {
```
agrojean-ledger, line 581: Can we have something in the `io` module that would work also for NBGL based apps ?

ledger_device_sdk/src/nbgl.rs
agrojean-ledger: Ok I can't reproduce the bugs I saw before, maybe I was not declaring the variables in the loop and it was causing issues.

agrojean-ledger:
@yogh333 You need to update the SDK version for crates.io deployment, don't you ?

agrojean-ledger (approved):

PR #160, opened 2024-05-31
Generate USB drivers bindings automatically

yogh333:


yogh333:
See https://github.com/LedgerHQ/app-radix-babylon/pull/14 for Radix

PR #164, opened 2024-06-17
Enable heap allocator in SDK and perform CString conversion for NBGL

yogh333:


yogh333:
To be tested with https://github.com/LedgerHQ/app-boilerplate-rust/pull/60

ledger_device_sdk/src/nbgl.rs
```diff
@@ -99,70 +106,20 @@ pub extern "C" fn io_recv_and_process_event() -> bool {
...
 
+    let setting_idx: usize = idx as usize;
+    SWITCH_ARRAY[setting_idx].initState = !SWITCH_ARRAY[setting_idx].initState;
+
     if let Some(data) = NVM_REF.as_mut() {
-        let setting_idx: usize = idx as usize;
         let mut switch_values: [u8; SETTINGS_SIZE] = data.get_ref().clone();
         switch_values[setting_idx] = !switch_values[setting_idx];
         data.update(&switch_values);
-        SWITCH_ARRAY[setting_idx].initState = switch_values[setting_idx] as nbgl_state_t;
```
agrojean-ledger, line 165: I'm not sure we want to separate the true value in the NVM from the SWITCH_ARRAY initState.

ledger_device_sdk/src/nbgl.rs
```diff
@@ -279,8 +200,36 @@ impl<'a> NbglHomeAndSettings<'a> {
...
+
+                let info_list: nbgl_contentInfoList_t = nbgl_contentInfoList_t {
+                    infoTypes: INFO_FIELDS.as_ptr() as *const *const c_char,
+                    infoContents: info_contents[1..].as_ptr() as *const *const c_char,
+                    nbInfos: INFO_FIELDS.len() as u8,
+                };
+
+                let icon: nbgl_icon_details_t = match self.glyph {
+                    Some(g) => g.into(),
+                    None => nbgl_icon_details_t::default(),
```
agrojean-ledger, line 218: It would be better I think to use NULL pointer if there is no icon to display... this is what the C api expects, not a structure with default values

agrojean-ledger: Why feel the need to rewrite my code like I didn't properly think about what I did ?

ledger_device_sdk/src/nbgl.rs
```diff
@@ -279,8 +200,36 @@ impl<'a> NbglHomeAndSettings<'a> {
         Reply: From<<T as TryFrom<ApduHeader>>::Error>,
     {
         unsafe {
-            if NVM_REF.is_some() {
```
agrojean-ledger, line 282: Also, I don't think we want to allow the settings to be displayed if there is no way to save them somewhere ? This can lead to some bug where the developer does not pass a reference to NVM but settings are still displayed but not working properly...

ledger_device_sdk/src/nbgl.rs
yogh333: Initial goal was to separate the settings set in RAM from the settings saved in NVRAM:

1. One can think about an app with settings but no persistent settings: after the app is restarted, the setting are reset. It does not prevent usage of NVRAM to save settings but we could also manage this use case: app with noi persistent settings.
2. Writing systematically in the flash every time a setting is modified can involve an flash wearing. That's why I think we should properly separate the saving of the settings.

ledger_device_sdk/src/nbgl.rs
yogh333: This is managed internally by Rust bindings:
```
impl Default for nbgl_icon_details_s {
    fn default() -> Self {
        let mut s = ::core::mem::MaybeUninit::<Self>::uninit();
        unsafe {
            ::core::ptr::write_bytes(s.as_mut_ptr(), 0, 1);
            s.assume_init()
        }
    }
}
```

ledger_device_sdk/src/nbgl.rs
yogh333: See my previous comments but if you think it is better to deal systematically with **persistent** settings ok

ledger_device_sdk/src/nbgl.rs
yogh333: Ok this has been reverted: when settings are defined, they are systematically saved as NVM data.

ledger_device_sdk/src/nbgl.rs
yogh333: ok

ledger_device_sdk/.cargo/config.toml
```diff
@@ -14,8 +14,8 @@ runner = "speculos --model stax --display=headless"
 runner = "speculos --model flex --display=headless"
 
 [unstable]
-build-std = ["core"]
+build-std = ["core", "alloc"]
 build-std-features = ["compiler-builtins-mem"]
 
 [build]
-target = "nanosplus"
+target = "stax"
```
agrojean-ledger, line 21: Why this change ? :thinking:

yogh333: This sets the default target to build for: https://doc.rust-lang.org/cargo/reference/config.html#buildtarget if we call `cargo build` without `--target`. Any target will fit, was more simple to use `stax` as the default one for my tests...

yogh333: And stax + 🦀  is the future 🚀

ledger_device_sdk/src/nbgl.rs
```diff
@@ -20,6 +22,11 @@ pub struct Field<'a> {
     pub value: &'a str,
 }
 
+pub struct CField {
```
agrojean-ledger, line 25: This doesn't need to be a public struct it seems we will only use it internally.

ledger_device_sdk/src/nbgl.rs
```diff
@@ -179,37 +135,17 @@ const INFO_FIELDS: [*const c_char; 2] = [
 pub struct NbglHomeAndSettings<'a> {
     glyph: Option<&'a NbglGlyph<'a>>,
     // app_name, version, author
-    info_contents: [*const c_char; 3],
-    settings_contents: nbgl_content_t,
+    info_contents: Vec<CString>,
```
agrojean-ledger, line 138: I don't think we need a Vec here as we only need 3 strings : version, developer, app_name.

ledger_device_sdk/src/nbgl.rs
```diff
@@ -279,8 +199,34 @@ impl<'a> NbglHomeAndSettings<'a> {
...
+                let icon: nbgl_icon_details_t = match self.glyph {
+                    Some(g) => g.into(),
+                    None => nbgl_icon_details_t::default(),
+                };
+
+                for (i, setting) in self.setting_contents.iter().enumerate() {
+                    SWITCH_ARRAY[i].text = setting[0].as_ptr();
+                    SWITCH_ARRAY[i].subText = setting[1].as_ptr();
+                    SWITCH_ARRAY[i].initState =
+                        NVM_REF.as_mut().unwrap().get_ref()[i] as nbgl_state_t;
```
agrojean-ledger, line 224: No check on NVM_REF as discussed yesterday ?

ledger_device_sdk/src/nbgl.rs
yogh333: Nowadays yes, but more future-proof if we use a `Vec`

ledger_device_sdk/src/nbgl.rs
yogh333: Not needed: as soon as some setting are defined through the `settings` call, NVM_REF is set.
So if no settings are defined, the `for` loop will be empty and NVM_REF not used...

ledger_device_sdk/src/nbgl.rs
```diff
@@ -227,49 +163,33 @@ impl<'a> NbglHomeAndSettings<'a> {
...
         self,
         nvm_data: &'a mut AtomicStorage<[u8; SETTINGS_SIZE]>,
-        settings_strings: &[[&str; 2]],
+        settings_strings: &[[&'a str; 2]],
     ) -> NbglHomeAndSettings<'a> {
         unsafe {
             NVM_REF = Some(transmute(nvm_data));
         }
 
-        if settings_strings.len() > SETTINGS_SIZE {
```
agrojean-ledger, line 249: I think we should put a restriction on how many settings_strings are allowed, it will lead to an app crash at line 220 if there are more `setting_strings` item than there are slots in `SWITCH_ARRAY`

agrojean-ledger: Maybe break the loop at line 220 if we reach max number of items in SWITCH_ARRAY, preventing a crash

yogh333: True, I have re-enabled the settings length check in the function to prevent any further crash.

ledger_device_sdk/src/nbgl.rs
```diff
@@ -162,7 +118,7 @@ unsafe fn settings_callback(token: ::core::ffi::c_int, _index: u8, _page: ::core
         let mut switch_values: [u8; SETTINGS_SIZE] = data.get_ref().clone();
         switch_values[setting_idx] = !switch_values[setting_idx];
         data.update(&switch_values);
-        SWITCH_ARRAY[setting_idx].initState = switch_values[setting_idx] as nbgl_state_t;
+        SWITCH_ARRAY[setting_idx].initState = !SWITCH_ARRAY[setting_idx].initState;
```
agrojean-ledger, line 121: I think we should leave ` SWITCH_ARRAY[setting_idx].initState = switch_values[setting_idx] as nbgl_state_t;` to make 100% sure there is no way for the displayed switch to have a value different than what is written in NVM

yogh333: done

agrojean-ledger (approved):

PR #165, opened 2024-06-18
WIP : add NBGL generic review.

agrojean-ledger:


ledger_device_sdk/src/nbgl.rs
```diff
@@ -473,6 +495,570 @@ impl<'a, const MAX_FIELD_NUMBER: usize, const STRING_BUFFER_SIZE: usize>
...
+    pub wrapping: bool,
+}
+
+/// Wrapper around the nbgl_contentTagValueList_t struct
+/// that contains the necessary variables used as pointers
+/// in the C struct.
+/// The variables are heap allocated and stored
+/// in the wrapper to ensure they live long enough.
+#[allow(dead_code)]
+struct TagValueListWrapper {
```
yogh333, line 725: I do not understand why have you defined all those wrappers ? And why storing all the variables in the heap ?

agrojean-ledger: Because the c_struct takes pointers to these variables... If I don't store them in the wrapper (as CString for example, which is why I mean by "on the heap"), then they don't live long enough... And the app crashes (or displays weird stuff)

agrojean-ledger: @yogh333 either we have wrappers to store variables or the API needs to ask for references (such as `&CStr`) to these variables. But we decided we would allow the user to use `&str` and not bother with `CString` and such.

ledger_secure_sdk_sys/build.rs
```diff
@@ -228,11 +228,12 @@ fn clone_sdk(device: &Device) -> PathBuf {
         ),
         Device::Flex => (
             Path::new("https://github.com/LedgerHQ/ledger-secure-sdk"),
-            "API_LEVEL_19",
+            "ux_sync_public_callbacks",
```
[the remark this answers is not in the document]
agrojean-ledger: Yes but... until I merge the ux_sync changes in the C SDK, we need to keep this branch (or another created on API_LEVEL_20

ledger_secure_sdk_sys/build.rs
```diff
@@ -346,10 +347,12 @@ impl SDKBuilder {
...
-        let sdk_path = match env::var("LEDGER_SDK_PATH") {
-            Err(_) => clone_sdk(&self.device),
-            Ok(path) => PathBuf::from(path),
-        };
+        // let sdk_path = match env::var("LEDGER_SDK_PATH") {
+        //     Err(_) => clone_sdk(&self.device),
+        //     Ok(path) => PathBuf::from(path),
+        // };
+
+        let sdk_path = clone_sdk(&self.device);
```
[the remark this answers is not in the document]
agrojean-ledger: Indeed

ledger_device_sdk/examples/nbgl_generic_review.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    let tag_values_list_2 = TagValueList {
+        pairs: &my_example_fields,
+        nb_max_lines_for_value: 2,
+        small_case_for_value: false,
+        wrapping: false,
+    };
+
+    let tag_value_confirm = TagValueConfirm {
+        tag_value_list: tag_values_list_2,
```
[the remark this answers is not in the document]
agrojean-ledger: TagValueConfirm takes ownership and consumes the tag_value_list_2 so you need another one. But in a real life scenario there shouldn't be too many cases where you need to deal with two of them anyway...

ledger_device_sdk/src/nbgl.rs
yogh333: Using heap is not mandatory I think. For instance, take a look on my [branch](https://github.com/LedgerHQ/ledger-device-rust-sdk/blob/568e3d88ee8326dee7992f0b6e79968b95910dec/ledger_device_sdk/src/nbgl.rs#L351C18-L351C32): `tag_value_list` is defined on the stack just before calling the NBGL C function `ux_sync_review`. As long as the scope of `tag_value_list` is not over before calling the C function, it is okay. Same for `Cstring`actually: if the CString lives on the stack and is in the same scope as the C NBGL function call, then there is no pb to pass the CString as a pointer (`CString.as_ptr()`): see for instance [here](https://github.com/LedgerHQ/ledger-device-rust-sdk/blob/568e3d88ee8326dee7992f0b6e79968b95910dec/ledger_device_sdk/src/nbgl.rs#L409).

agrojean-ledger: @yogh333 of course it works in the simple case of a CString (which by the way is allocated on the heap anyway even if you declare it locally in a function, like String), because it lives long enough in the function `show` : 

* function show is called,
* CString is allocated,
* C Api function is called with pointer to CString,
* C function returns
* show function returns, CString is dropped

In my case the sequence is a lot different : 

* `addcontent` of the generic review is called with a content struct such as  `TagValueList` as parameter. The `addcontent` function takes ownership of the `TagValueList` variable and converts it to the wrapper structure with the `impl<'a> From<TagValueList<'a>> for TagValueListWrapper` trait function.
* Variables are created (such as `let mut c_field_names: Vec<CString> = Vec::new();`) to store the different field names to be displayed. We actually don't care they are on the heap, the important thing is they are stored in the wrapper.
* Pointers to these variables are used to intialize the fields of the underlying C struct.
* The `From` trait function **returns** the wrapper which contains the C struct and the actual variables it points to, ensuring they live long enough. The original `TagValueList` (for instance) is dropped.
* The wrapper is added to the Vec in the NbglGenericReview struct, which can then use this list to display the different screens when the show function is called.

We are not in the case of a single `show` function called, we need `add_content` + `From` trait function to perform conversions

If you want to discuss a better implementation I'd very much appreciate if we do it *before*  we merge the PR. I wouldn't like if the next day you rewrite my code because it doesn't suit you.

agrojean-ledger: @yogh333 If we use a trait function to convert from let's say `TagValueList` to a `nbgl_contentTagValueList_t` (the C struct) directly, we can declare variables and use the pointers to these variables in the `nbgl_contentTagValueList_t`, but once the conversion function returns, these variables have nowhere to go and will be dropped...

I tried.

agrojean-ledger: I will update the code comments to avoid mentioning `heap` everywhere. We actually don't care it's on the heap. The important thing is we need to ensure these variables live long enough to be used when we actually call the C api.

yogh333: @agrojean-ledger Thanks for your explanation. However, we must challenge the need to store **all** the contents before calling `show`. Why not calling `show` for each content, and thus giving up the `add_content` function ? In that case, a conversion from `TagValueList` to `nbgl_contentTagValueList_t ` in the show function is enough.

agrojean-ledger: @yogh333 the C generic review API works by passing a `contentsList: *const nbgl_content_t,`, which has to contain all the different content structures... 

There's a possibility to avoid having to preallocate everything by using the `contentGetterCallback`, when the app does not have knowledge of all the contents before calling the function. 

The API uses a union to allow both behaviors : 

```
pub union nbgl_genericContents_t__bindgen_ty_1 {
    pub contentsList: *const nbgl_content_t,
    pub contentGetterCallback: nbgl_contentCallback_t,
}
```

But as its name implies, the second option uses a *callback*. I didn't put too much effort looking at this one because we know : 

1. Using callbacks is a pain (may not even be feasible with the current C ux_sync library),
2. It's another kind of usage (where the app does not have all the content ready before displaying the review)

If you want a "typical" example of generic review usage, take a look at the ethereum app : https://github.com/LedgerHQ/app-ethereum/blob/76e8690a68fb84910d09fbfd14cc975cefabe41a/src_nbgl/ui_display_privacy.c#L23
You can see, everything is allocated before calling the generic review API.

agrojean-ledger: @yogh333 I will try with the callback.

yogh333: Sorry but I think you miss my point. After looking into the C API, [nbgl_useCaseGenericReview](https://github.com/LedgerHQ/ledger-secure-sdk/blob/1aa3c0a56c748dc6085afdee2d489a66257d5ee7/lib_nbgl/src/nbgl_use_case.c#L2621) uses `nbgl_genericContents_t`:
```
typedef struct {
    bool callbackCallNeeded;  ///< indicates whether contents should be retrieved using
                              ///< contentsList or contentGetterCallback
    union {
        const nbgl_content_t *contentsList;  ///< array of nbgl_content_t (nbContents items).
        nbgl_contentCallback_t
            contentGetterCallback;  ///< function to call to retrieve a given content
    };
    uint8_t nbContents;  ///< number of contents
} nbgl_genericContents_t;
```
In this struct, we can use `nbgl_content_t` which needs `nbgl_contentTagValueList_t`. So I think we can intialize all those C structs just before calling the C function `nbgl_useCaseGenericReview`, i.e when calling the Rust `show` function. To me there is no need to use a  `TagValueListWrapper`. Let's assume we keep the `add_content` function: `add_content` function could store all the contents using Rust types (`String`, `&str`) but the final conversion is directly performed when calling the `show` function, right before calling the `nbgl_useCaseGenericReview`. One might also think about the Rust `show` function that passes **all** the content (using Rust types/structures) to be displayed, and once again all the conversion are performed just before calling `nbgl_useCaseGenericReview`. Hope it clarifies my proposal 🙏  (and yes callback is a pain, hope I have managed to save you from that hell 😉 )

ledger_device_sdk/examples/nbgl_generic_review.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    let mut review: NbglGenericReview = NbglGenericReview::new()
+        .add_content(NbglPageContent::CenteredInfo(centered_info))
+        .add_content(NbglPageContent::InfoButton(info_button))
+        .add_content(NbglPageContent::InfoLongPress(info_long_press))
+        .add_content(NbglPageContent::TagValueList(tag_values_list))
+        .add_content(NbglPageContent::TagValueConfirm(tag_value_confirm))
+        .add_content(NbglPageContent::InfosList(infos_list));
+
+    review.show("Reject Example", "Example Confirmed", "Example Rejected");
```
kingofpayne, line 89: For the demonstration, the result of this call should be used to take an action if the user validated (or not).

agrojean-ledger: Ok, I'll display a status

ledger_device_sdk/src/nbgl.rs
```diff
@@ -371,6 +375,446 @@ impl<'a> NbglReview<'a> {
...
+            let value = CString::new(field.value).unwrap();
+            let tag_value = nbgl_contentTagValue_t {
+                item: name.as_ptr() as *const c_char,
+                value: value.as_ptr() as *const c_char,
+                ..Default::default()
+            };
+            c_field_strings.push(tag_value);
+            c_field_names.push(name);
+            c_field_values.push(value);
+        }
```
kingofpayne, line 517: As we know in advance the number of items that `c_field_strings`, `c_field_names` and `c_field_values` will have, it is better to initialize the vectors using [Vec::with_capacity](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity) . This will reduce the number of allocations/frees made in the heap.

kingofpayne: ... or you can use `collect` for each vector, which should be optimized.

agrojean-ledger: I have used `with_capacity`

kingofpayne (approved):

PR #170, opened 2024-07-08
Add streaming review support in NBGL module

agrojean-ledger:
ExampleStreamingReview


ledger_device_sdk/src/nbgl.rs
```diff
@@ -821,6 +865,116 @@ impl NbglGenericReview {
     }
 }
 
+pub struct NbglStreamingReview {
+    icon: nbgl_icon_details_t,
+    tx_type: TransactionType,
+}
+
+impl NbglStreamingReview {
+    pub fn new(tx_type: TransactionType) -> NbglStreamingReview {
```
agrojean-ledger, line 874: @yogh333 I have added the transaction type parameter to the `new` method. it's saved in the structure and used in various places (`start`, `continue_review` and `finish`)

I didn't want to have a separate function to set it.

What do you think ? Should I add this parameter to the other review structures as required ?

yogh333: Yes ok for me

yogh333: But I am wondering if it can be also generalised to other structs to provide constant strings when calling `new()`, e.g.:

- `NbglReview::new(title, subtitle, finish_title)`
- `NbglHomeAndSettings::new(info_contents)`
- `NbgAddressReview::new(verify_string)`

agrojean-ledger: I would prefer keeping the "builder pattern" style functions that seem to allow greater flexibility in using the structures of our API. Athough it's true that I'm not sure we would call `new` and then **not** call `titles` (for the `NbglReview`)

yogh333 (changes requested):
LGTM but `ledger_secure_sdk_sys` version should be increased

agrojean-ledger:
> LGTM but `ledger_secure_sdk_sys` version should be increased

Done.

yogh333 (approved):

PR #173, opened 2024-07-26
Nbgl add use case choice

agrojean-ledger:
* Add "choice" use case in NBGL module.
* Implement warning screens for reviewing transactions in "blind" mode, directly into the NBGL module (for `NbglReview` and `NbglStreamingReview` when `blind()` is called before `show()`)
* Code cleanup, remove the name of the crate `ledger_secure_sdk_sys` in NBGL when calling stuff from it, it's useless as we import everything from it.


yogh333 (approved):
LGTM !

PR #177, opened 2024-07-31
Remove status screens at the end of reviews, add new structures

agrojean-ledger:
so that developers may add such screens themselves in the apps.


yogh333 (approved):
LGTM

PR #178, opened 2024-08-08
updates `HEAP_SIZE` to be configurable for `embedded alloc`

jowparks:
This should be optionally configurable as it is possible to use more than 8KB memory on device.

### DIFF
--- a/ledger_secure_sdk_sys/build.rs
+++ b/ledger_secure_sdk_sys/build.rs
@@ -1,5 +1,6 @@
 extern crate cc;
 use glob::glob;
+use std::fs;
 use std::path::{Path, PathBuf};
 use std::process::Command;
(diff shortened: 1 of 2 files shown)


yogh333:
@jowparks  Hi, thanks for your PR. Could you please rebase ? Additionally, I would recommend to retrieve the HEAP_SIZE from the app `config.toml` file (see [here](https://doc.rust-lang.org/cargo/reference/config.html#env)) and using it directly in `lib.rs` without using the temp file `heap_size.rs` if it is feasible (I think so...). Thank you !

yogh333:
@jowparks Could you please rebase your branch ? After a deepper review, it looks that it can be merged into our SDK. Thanks !

yogh333:
Close this one as #183 duplicates it

PR #179, opened 2024-08-16
Support HMAC algorithms

bboilot-ledger:
Hello :wave: 

This PR adds the support of HMAC for Ripmd160, sha256 and sha512 to the SDK with a similar implementation as what has been done for hashes algorithm.

This will be needed for https://github.com/LedgerHQ/vanadium/issues/2

Few things to note:
- ~~`sha224` and `sha384` are not supported because the C SDK does not implement the related `*_init_no_throw` methods. I have no idea on how to handle the functions that throw from rust plus the macro would not be generic anymore.~~
- I did not bump the SDK minor version since I don't know the current versioning flow.

Let me know if any change is needed.


ledger_device_sdk/src/hmac.rs
```diff
@@ -0,0 +1,198 @@
...
+                unsafe { mem::transmute::<&mut $ctxname, &mut cx_hmac_t>(&mut self.ctx) }
+            }
+
+            fn as_ctx(&self) -> &cx_hmac_t {
+                unsafe { mem::transmute::<&$ctxname, &cx_hmac_t>(&self.ctx) }
+            }
+
+            fn new(key: &[u8]) -> Self {
+                let mut ctx: $typename = Default::default();
+                let _err = unsafe { $initfname(&mut ctx.ctx, key.as_ptr(), key.len() as u32) };
```
yogh333, line 119: @bboilot-ledger I am not sure to understand why two different versions of `impl_hmac` are required. In the C SDK, `no_throw` suffixed functions take a `size_t` as key length parameter whereas `throw` functions take an `unsigned int` key length parameter. On a 32-bit platform, both are 32 bits so passing key.len() (usize) should be ok ? Or am I missing anything 🤔 ?

bboilot-ledger: I completely agree with you. But the compiler is not really happy about that:
image
(same result if I try the opposite way).

I did try to check dynamically the function name and cast `key.len()` to the right type accordingly (which avoids code duplication) but still, it won't compile because the check is done at run time. I did not find a way to do "conditional" macros.
I may be missing something there (I am quite new on Rust), so if you see a way of avoiding code duplication, I will be super happy to implement it!

yogh333: Looks like using `try_into()` enables to avoid code duplication:
````
fn new(key: &[u8]) -> Self {
    let mut ctx: $typename = Default::default();
    let _err = unsafe {
        $initfname(&mut ctx.ctx, key.as_ptr(), key.len().try_into().unwrap())
    };
    ctx
}
````

yogh333 (approved):

ledger_device_sdk/src/hmac.rs
bboilot-ledger: Oh nice, I did not think about that. Thanks for the help @yogh333 !

PR #180, opened 2024-08-19
Allow heap feature to be disabled.

ryankurte:
A while back the SDK introduced a heap allocator for everything but the `nanos`, which uses a config trick to enable this for all other platforms, however, the large size chosen for this default heap may exceed the available memory for applications that are already near RAM limits, and the target specific inclusion trick used to select this does not allow the heap feature to be disabled.

This PR simplifies the behaviour of the `heap` feature so this has no effect on the unsupported `nanos` target and can be disabled for apps that need to control the size of the allocated heap.

### DIFF
--- a/ledger_device_sdk/Cargo.toml
+++ b/ledger_device_sdk/Cargo.toml
@@ -21,13 +21,11 @@ rand_core = { version = "0.6.3", default_features = false }
 zeroize = { version = "1.6.0", default_features = false }
 numtoa = "0.2.4"
 const-zero = "0.1.1"
-
-[target.'cfg(target_os="nanos")'.dependencies]
-ledger_secure_sdk_sys = {path = "../ledger_secure_sdk_sys", version = "1.4.3"}
-
-[target.'cfg(not(target_os="nanos"))'.dependencies]
-ledger_secure_sdk_sys = {path = "../ledger_secure_sdk_sys", version = "1.4.3", features = ["heap"]}
+ledger_secure_sdk_sys = {path = "../ledger_secure_sdk_sys", version = "1.4.3", default_features = false}
 
 [features]
 speculos = []
 ccid = []
+heap = [ "ledger_secure_sdk_sys/heap" ]
+
+default = [ "heap" ]
--- a/ledger_device_sdk/Cargo.toml
+++ b/ledger_device_sdk/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ledger_device_sdk"
-version = "1.14.0"
+version = "1.14.1"
 authors = ["yhql", "yogh333", "agrojean-ledger", "kingofpayne"]
 edition = "2021"
 license.workspace = true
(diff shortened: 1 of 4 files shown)


ledger_secure_sdk_sys/src/lib.rs
```diff
@@ -61,15 +61,15 @@ unsafe impl critical_section::Impl for CriticalSection {
 /// The heap is stored in the stack, and has a fixed size.
 /// This method is called just before [sample_main].
 #[no_mangle]
-#[cfg(feature = "heap")]
+#[cfg(all(feature = "heap", not(target_os = "nanos")))]
 extern "C" fn heap_init() {
-    const HEAP_SIZE: usize = 8192;
+    const HEAP_SIZE: usize = 1024;
```
kingofpayne, line 66: This change is not part of the PR description, maybe this has been committed by mistake?
I think this should be reverted. With 8K heap by default, there is still plenty of stack available.
Also, this setting may be tunable in the future (see #178)

ledger_device_sdk/Cargo.toml
```diff
@@ -21,13 +21,11 @@ rand_core = { version = "0.6.3", default_features = false }
 zeroize = { version = "1.6.0", default_features = false }
 numtoa = "0.2.4"
 const-zero = "0.1.1"
-
-[target.'cfg(target_os="nanos")'.dependencies]
-ledger_secure_sdk_sys = {path = "../ledger_secure_sdk_sys", version = "1.4.3"}
-
-[target.'cfg(not(target_os="nanos"))'.dependencies]
-ledger_secure_sdk_sys = {path = "../ledger_secure_sdk_sys", version = "1.4.3", features = ["heap"]}
+ledger_secure_sdk_sys = {path = "../ledger_secure_sdk_sys", version = "1.4.3", default_features = false}
```
kingofpayne, line 24: `heap` feature is not enabled by default in the `ledger_secure_sdk_sys`. It think we can omit `default_features = false` in that dependency declaration. Otherwise, if in the future a new feature is added by default to `ledger_secure_sdk_sys`, it would be disabled by this statement.

ledger_secure_sdk_sys/src/lib.rs
ryankurte: oops yep that's a mistake!

yogh333 (changes requested):
LGTM but both ledger_device_sdk and ledger_secure_sdk_sys crates shall be version bumped (`1.14.1` for ledger_device_sdk and `1.4.4` for ledger_secure_sdk_sys)

agrojean-ledger:
@ryankurte @yogh333 @kingofpayne we need to be careful in allowing the heap to be disabled : the NBGL module relies on it for the `alloc` crate usage.

Apps for `Stax` and `Flex` that use NBGL (so most likely 100% of them) will not compile without the heap allocator enabled.

yogh333:
@agrojean-ledger Yes you are right. This PR makes the heap feature enabled by default except on nanos so there is no issue on my side. To be disabled the `--no-default-features` shall be used explicitly. And when compiled for Nano S (no NBGL support), the heap feature is disabled automatically. So it is ok for me.

yogh333 (approved):

agrojean-ledger:
> @agrojean-ledger Yes you are right. This PR makes the heap feature enabled by default except on nanos so there is no issue on my side. To be disabled the `--no-default-features` shall be used explicitly. And when compiled for Nano S (no NBGL support), the heap feature is disabled automatically. So it is ok for me.

Ok then, up to developers to be careful when disabling the feature for a specific target.

agrojean-ledger (approved):

PR #186, opened 2024-09-17
Remove warning message when performing blind signing (NBGL)

yogh333:


tdejoigny-ledger:
@yogh333 if a message is blind-signed we need to use (now) the new blind-signing flow.

agrojean-ledger:
@yogh333 why do you want to remove it ? Either we update it or we use the new API from de C SDK to provide warnings

yogh333:
Updated following your comments

PR #188, opened 2024-09-20
Y333/new nbgl

yogh333:


ledger_device_sdk/src/nbgl/nbgl_home_and_settings.rs
```diff
@@ -190,4 +203,65 @@ impl<'a> NbglHomeAndSettings {
             }
         }
     }
+
+    /// Show the home screen and settings page.
+    /// This function returns immediately after the screen is displayed.
+    pub fn display(&mut self) {
```
agrojean-ledger, line 209: @yogh333 could be called `show_and_return` it would outline the difference with `show` a bit better imo.

tdejoigny-ledger (approved):

PR #193, opened 2024-09-27
Some new features

emmanuelm41:
- allow to choose light review or not  
- create new page style

### DIFF
--- a/ledger_device_sdk/src/nbgl/nbgl_review.rs
+++ b/ledger_device_sdk/src/nbgl/nbgl_review.rs
@@ -9,6 +9,7 @@ pub struct NbglReview<'a> {
     glyph: Option<&'a NbglGlyph<'a>>,
     tx_type: TransactionType,
     blind: bool,
+    light: bool,
 }
 
 impl SyncNBGL for NbglReview<'_> {}
@@ -22,6 +23,7 @@ impl<'a> NbglReview<'a> {
             glyph: None,
             tx_type: TransactionType::Transaction,
             blind: false,
+            light: false,
         }
     }
 
@@ -32,6 +34,13 @@ impl<'a> NbglReview<'a> {
     pub fn blind(self) -> NbglReview<'a> {
         NbglReview {
             blind: true,
+            ..self
+        }
+    }
+
+    pub fn light(self) -> NbglReview<'a> {
+        NbglReview {
+            light: true,
             ..self
         }
     }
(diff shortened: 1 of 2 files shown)


ledger_secure_sdk_sys/build.rs
```diff
@@ -209,26 +209,31 @@ fn retrieve_target_file_infos(
 
 /// Fetch the appropriate C SDK to build
 fn clone_sdk(device: &Device) -> PathBuf {
-    let (repo_url, sdk_branch) = match device {
+    let (repo_url, sdk_branch, commit_hash) = match device {
         Device::NanoS => (
             Path::new("https://github.com/LedgerHQ/ledger-secure-sdk"),
             "API_LEVEL_LNS",
+            "93fa69b170dd819c6b93d5cf49dd21acee51ffd0"
```
yogh333: Cloning the C-SDK only occurs when working out of our docker container (see [here](https://github.com/LedgerHQ/ledger-app-builder)) or when the LEDGER_SDK_PATH is not defined: cargo-ledger sets this environment variable [automatically](https://github.com/LedgerHQ/ledger-device-rust-sdk/blob/8d19e155b84af5876bc0383d02b00dcf2eeb3f9e/cargo-ledger/src/main.rs#L223)). Anyway, it is recommended to work with head of API_LEVEL_XX branch in order to have the most up to date C-SDK, and this is what we enforce when deploying an app. Moreover, in case of any issue (C-SDK compilation error or Rust binding error etc...), it enables to detect it and to fix the Rust SDK as soon as possible.

ledger_secure_sdk_sys/build.rs
```diff
@@ -209,26 +209,31 @@ fn retrieve_target_file_infos(
 
 /// Fetch the appropriate C SDK to build
 fn clone_sdk(device: &Device) -> PathBuf {
-    let (repo_url, sdk_branch) = match device {
+    let (repo_url, sdk_branch, commit_hash) = match device {
         Device::NanoS => (
             Path::new("https://github.com/LedgerHQ/ledger-secure-sdk"),
             "API_LEVEL_LNS",
+            "93fa69b170dd819c6b93d5cf49dd21acee51ffd0",
```
yogh333, line 216: We want to use the head of the branch when cloning the C-SDK (this is what is done when deploying an app for instance).

ledger_secure_sdk_sys/build.rs
emmanuelm41: Thanks yogh for your reply. I understand this is a bit controversial, and I understand your point. At the same time, if the actual reference to a commit hash is not set, any given version of the rust sdk can suffer from non backward compatible changes on the C SDK. I was working on a given version without the docker image, and suddenly the app was not compiling anymore because a c file was removed on the C SDK. As there is no reference to the C SDK version used or "valid" at a given time, it is impossible to easily know which C SDK version it was valid with.

yogh333: @emmanuelm41 I recommend to use our Docker images to avoid such troubles: our Rust Docker containers (builder and dev-tools) embed a specific version of the C-SDK and they are updated when a new reference C-SDK is published. Or if you prefer working out of Docker,  you can clone the C-SDK by yourself, checking out the proper commits i.e. the ones used in our Dockers (check [here for the ref tags](https://github.com/LedgerHQ/ledger-app-builder/blob/9e14b65b24a918fed7e9a4a640b6f0116baa2c30/lite/Dockerfile#L86)) then setting LEDGER_SDK_PATH properly and that's it. Last but not least, the issue you have encountered is pretty rare and has been fixed in the last version of the Rust SDK, it should not occur every days.

yogh333 (changes requested):
LGTM except the modifications in build.rs

ledger_secure_sdk_sys/build.rs
emmanuelm41: Thanks yogh! I will do that. Commit dropped

yogh333 (approved):

PR #198, opened 2024-10-07
feat: add missing app os flags for appInfo instruction

emmanuelm41:
In the C SDK, the appInfo instruction has the appFlags available. On the Rust bindings, that is missing. That makes the same command for two different apps incompatible. I tested it locally, and it is possible to add it. I am not aware of reasons why it was missing. I suppose it was forgotten. 

Thanks!


tdejoigny-ledger:
cc: @agrojean-ledger @yogh333

NullSoldier:
Thank you, we use this in our ledger rust app, and the crash from trying to read past the end of the buffer because flags are not sent is causing our app some problems. We really appreciate this and look forward to getting this merged.

yogh333 (approved):
LGTM
Thanks !

PR #199, opened 2024-10-07
feat: use correct status word when device is locked

emmanuelm41:
When the app is active, and the device gets locked, the device returns "Instruction Not Supported". It does not seem correct, as the instruction is indeed supported. The device is actually locked instead. I changed the status word used on this scenario, in order to inform properly to the client.

### DIFF
--- a/ledger_device_sdk/src/io.rs
+++ b/ledger_device_sdk/src/io.rs
@@ -23,6 +23,7 @@ pub enum StatusWords {
     UserCancelled = 0x6e04,
     Unknown = 0x6d00,
     Panic = 0xe000,
+    DeviceLocked = 0x5515,
 }
 
 #[derive(Debug)]
--- a/ledger_device_sdk/src/ui/gadgets.rs
+++ b/ledger_device_sdk/src/ui/gadgets.rs
@@ -603,7 +603,8 @@ impl<'a> MultiPageMenu<'a> {
                         // pin lock management
                         let (_res, ins) = UxEvent::block_and_get_event::<Temp>(self.comm);
                         if let Some(_e) = ins {
-                            self.comm.reply::<io::StatusWords>(io::StatusWords::Unknown);
+                            self.comm
+                                .reply::<io::StatusWords>(io::StatusWords::DeviceLocked);
                         }
                         // notify Ticker event only when redisplay is required
                         return EventOrPageIndex::Event(io::Event::Ticker);


NullSoldier:
Thank you, we actually intercept and handle this incorrect status word to know when the app is locked.

https://github.com/iron-fish/ironfish/blob/staging/ironfish-cli/src/ledger/ledger.ts#L44-L49

This is a great fix, looking forward to getting this merged and released.

ledger_device_sdk/src/io.rs
```diff
@@ -23,6 +23,7 @@ pub enum StatusWords {
     UserCancelled = 0x6e04,
     Unknown = 0x6d00,
     Panic = 0xe000,
+    DeviceLocked = 0x5515,
```
yogh333, line 26: Maybe this should be 0x6e05 ?

yogh333 (changes requested):

ledger_device_sdk/src/io.rs
emmanuelm41: On our JS lib, 0x5515 is what we understand as LockedDevice. Why do you think it should be 0x6e05? This is not an app error, but more like a device/OS level error. What do you think?

yogh333: Yes indeed you are right, it is not an app error. Which JS lib are you talking about ? I am ok to define an new StatusWord but why 0x5515 ? We have also SyscallError (`0x68xx`) but it is different. Is there a group of errors that could be returned as `0x55xx`?

emmanuelm41: This is the ledger C SDK.  
image

emmanuelm41: That is the error you get if you try to interact with an app while the pin lock is enabled.

emmanuelm41: Source:
https://github.com/LedgerHQ/ledger-secure-sdk/blob/ddfd051f46f43e48a92dc6940eebd43d77aa0e42/include/errors.h#L126

emmanuelm41: Some more context 
image

yogh333: Ok good, let's use it then !

yogh333 (approved):

yogh333:
@emmanuelm41 Sorry, but could you please revert the last commit (merge master branch) + rebase your PR 🙏 ?
Thanks !

emmanuelm41:
@yogh333 rebased!

PR #200, opened 2024-10-10
Assert Heap Size 

yogh333:


emmanuelm41:
Hey @yogh333! Any reason why this check was added? It is quite inconvenient in cases where there is a limited size between stack and heap. Between 8192, and 16384 there is a big gap, and could potentially block apps where the stack is important too.

emmanuelm41:
Particularly, this could break the Ironfish DKG app that we are currently reviewing on ledger side.

emmanuelm41:
This [PR](https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/208/files) increases the stack usage on 359 bytes (because of the warning icon), and that breaks our app. We cannot decrease the heap size a bit, and 8192 is not enough. :/

yogh333:
@emmanuelm41 You are right, I have modified the assert in this PR #209

emmanuelm41:
@yogh333 thanks a lot for the quick fix. I think it will work now!! Again, thanks!

PR #201, opened 2024-10-10
Make link-wrap.sh Linux and macOS cross-platform compatible

harrychin:
Replacing `stat -c %z` with `wc -c < ' lets link-wrap.sh work on both Linux and macOS, whereas the stat command works differently between the two platforms.


harrychin:
Hi!

Just a tiny quality of life PR so the tiny corner of macOS devs don't need to keep patching link-wrap.sh 😄

PR #202, opened 2024-10-15
Display the s_arrow properly

Lbqds:
This PR improves the display of the `s_arrow` when switching between the field pages and the confirm/cancel page in `MultiFieldReview`.


yogh333:
@Lbqds BAGL library is about to be replaced by NBGL on all devices (`Nano S+`, `Nano X`, `Stax` and `Flex`). So sorry we do not integrate any new fix in BAGL gadgets.

PR #203, opened 2024-10-15
Make link-wrap.sh Linux and macOS cross-platform compatible

harrychin:
Replace `stat -c %z` with `wc -c <` as the native stat command for macOS doesn't support that option, but wc is supported similarly on both platforms.


yogh333 (changes requested):
@harrychin LGTM but could you please rebase your branch 🙏? Thank you !

Issue #214, opened 2024-11-05
cargo ledger build failed due to a newer version cargo.lock file


Pana:
More and more Rust crates are starting to use version 4 format of Cargo.lock. For example, `ethereum-types`, a foundational library for Ethereum application development by ParityTech, leads to compilation failures when introduced.

```sh
Use Rust nightly toolchain: nightly-2023-11-10
error: failed to parse lock file at: /app/Cargo.lock

Caused by:
  lock file version 4 requires `-Znext-lockfile-bump`
thread 'main' panicked at /opt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-ledger-1.5.1/src/main.rs:274:41:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Run the building command with one additional option `-Znext-lockfile-bump` can solve it, but don't know how to set it when using VS code extension.

It's recommend to upgrade the build image's rust version to a newer one to solve it, for example rustc 1.79

yogh333:
Yes Rust version to be upgraded before EOY 🤞

PR #225, opened 2024-12-13
Swap support

yogh333:


yogh333:
See #214 regarding the Cargo.lock version => Rust toolchain to be upgraded

PR #230, opened 2025-01-10
BT receive buffer is external

yogh333:


yogh333:
Unneeded fix, close without merging

PR #231, opened 2025-01-14
Increase SPI buffer size

yogh333:


yogh333:
Should be merged asap to fix issue with Near app

ledger_device_sdk/src/uxapp.rs
tdejoigny-ledger, line 1: line 58: 256 instead of 128 ? ---> `let mut spi_buffer = [0u8; 128];`

yogh333: Good catch

tdejoigny-ledger (approved):

PR #233, opened 2025-01-15
Swap review

yogh333:


ledger_device_sdk/src/libcall/swap.rs
```diff
@@ -221,21 +229,21 @@ pub enum SwapResult<'a> {
...
-            SwapResult::CheckAddressResult(&mut ref p, r) => {
-                *(p.result) = r;
-            }
-            SwapResult::PrintableAmountResult(&mut ref p, s) => {
+    match res {
+        SwapResult::CheckAddressResult(&mut ref p, r) => {
+            unsafe { *(p.result) = r };
+        }
+        SwapResult::PrintableAmountResult(&mut ref p, s) => {
+            if s.len() < (MAX_PRINTABLE_AMOUNT_SIZE - 1).try_into().unwrap() {
```
bboilot-ledger, line 237: In case of error, amount_str must be empty:
```C
/* Format printable amount including the ticker from specified parameters.
 *
 * Must set empty printable_amount on error, printable amount otherwise */
void swap_handle_get_printable_amount(get_printable_amount_parameters_t *params);
```

I am unsure if we should assume that it will be zeroed by default. Wdyt?

bboilot-ledger (commented):
Only one last point on my side, otherwise it looks good to me!

ledger_device_sdk/src/libcall/swap.rs
yogh333: Agreed

bboilot-ledger (approved):

PR #235, opened 2025-01-23
Generate glyphs automatically when building sys crate

yogh333:


ledger_secure_sdk_sys/build.rs
```diff
@@ -400,6 +400,59 @@ impl SDKBuilder {
...
+
+        for folder in glyph_folders.iter() {
+            for file in std::fs::read_dir(folder).unwrap() {
+                let path = file.unwrap().path();
+                let path_str = path.to_str().unwrap().to_string();
+                cmd.arg(path_str);
+            }
+        }
+
+        let _ = cmd.output();
```
tdejoigny-ledger, line 453: could be useful to return an error if command failed ?

tdejoigny-ledger (approved):
At some point, it would be good to define configuration arrays to avoid having to change multiple parts of the code when adding a new device.

yogh333:
> At some point, it would be good to define configuration arrays to avoid having to change multiple parts of the code when adding a new device.

Agreed. Should be done when migrating Nanos from BAGL to NBGL I think (good opportunity).

ledger_secure_sdk_sys/build.rs
yogh333: Yes indeed. But nowadays, there is no error management in this build script (see `main()`). Should be also considered when refactoring this file.

PR #239, opened 2025-02-04
Fix pin lock issue during io reset + Fix S+ and X SRAM size

yogh333:
Fix from https://ledgerhq.atlassian.net/browse/B2CA-1967

### DIFF
--- a/ledger_secure_sdk_sys/src/c/src.c
+++ b/ledger_secure_sdk_sys/src/c/src.c
@@ -266,6 +266,8 @@ void c_reset_bss() {
   memset(bss, 0, bss_len);
 }
 
+bolos_ux_params_t G_ux_params;
+
 void c_boot_std() {
     // below is a 'manual' implementation of `io_seproxyhal_init`
 #ifdef HAVE_MCU_PROTECT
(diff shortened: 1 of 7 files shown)


ledger_secure_sdk_sys/src/c/src.c
```diff
@@ -277,6 +279,21 @@ void c_boot_std() {
...
 #endif
 
+#ifndef TARGET_NANOS
+    // Warn UX layer of io reset to avoid unwanted pin lock
+    memset(&G_ux_params, 0, sizeof(G_ux_params));
+    G_ux_params.ux_id = BOLOS_UX_IO_RESET;
+
+    // If the app has just been booted from the UX, multiple os_ux calls may be necessary
+    // to ensure UX layer has take the BOLOS_UX_IO_RESET instruction into account.
+    for (uint8_t i = 0; i < 2; i++) {
```
fbeutin-ledger, line 289: Why 2 ?

yogh333: Good question, just a copy/paste from the C SDK: https://github.com/LedgerHQ/ledger-secure-sdk/blob/55afb4121b299635fae561239e44d2d5a60e3780/src/os_io_seproxyhal.c#L428

PR #241, opened 2025-02-10
Y333/nbgl support for nanos

yogh333:


yogh333:
Sorry but I have to force the merge. CI failures are due to invalid Docker builders. They have to be updated but htis is possible only after merging this PR and publishing the new cargo-ledger crate.

PR #258, opened 2025-04-21
feat: Add const generic address buffer size to Swap params

Fingolfin69:
Adds `const` generic address buffer size to Swap parameter structures


yogh333 (commented):
LGTM, please bump SDK version, thank you !

PR #267, opened 2025-07-01
Forward all events to OS by default

yogh333:
Thanks @sportron-ledger for your support


ledger_device_sdk/src/seph.rs
sportron-ledger, line 1: As all events are forwarded and status not handled by the SDK, the following modification might be unecessary.

PR #295, opened 2025-09-19
New IO module

bigspider:
Adds a new (alternative) `io` module that can be opted in by enabling the `io_new` feature.

The goals of the new io modules are the following:
a) Allow arbitrary size for the IO buffer, while defaulting to the usual size
b) Avoid memory waste; a single buffer used as a duplex channel is enough
c) Provide a cleaner abstraction for the duplex channel, avoiding the possibility of mistakenly using the buffer simultaneously to read from a command/apdu, or write to it for the response.

Goals (a) and (b) are fully achieved.
Goal (c) is achieved for the default usage flow as demonstrated in the boilerplate below; however, the API is not very strict so it does technically allow to make mistakes (like forgetting to respond altogether). It might be possible to come up with even stricter APIs that prevent even more misuse of the communication channel; this is left for future exploration.

In `io_new`, data is _never_ copied implicitly to or from the IO buffer. Rust ownership guarantees that the buffer is always only owned in a single place, so it can't be in 'read mode' and in 'write mode' at the same time.

In general, I think the new API is much cleaner, modular and easier to extend with future more complex features.

## Compatibility

`io` is renamed to `io_legacy`, and the new `io_new` module is part of the crate. However, they are not public modules.

If the `io_feature` is not enabled (current default), every member of `io_legacy` is exported under `io`; otherwise, `io_new` is the `io` module.

This allows to keep the new module entirely opt-in, and if the new module is considered better than the old one, migration of existing apps can happen over time.

## Breaking changes
 
As `io_new` is opt-in, there is no breaking change for existing apps. The only exception is that `Comm::new()` is no longer `const`, because that would prevent the refactoring for callbacks (see below). No app seems to be affected by this.

The `init_comm` is now obsolete, as callbacks are initialized in the constructor of the `Comm` object, which should generally be a singleton. Therefore, `init_comm` is now empty (kept only to prevent a breaking change).

## Feature parity

New `io_new` module should already have nearly all the features of the old module, including the BOLOS apdu handling and the `expected_cla` feature.

The only exception should be the 'automatic wait for a ticker event during apdu_send` of `io_legacy`, which is a workaround for specific apps that has a substantial performance impact in communication.

This could be reintroduced as an opt-in feature explicitly requested by apps, unless a better solution that avoids the performance impact is found.

## NBGL callbacks (refactoring)
The PR also includes a big refactoring of the NBGL integration with the `Comm<>` object, using callbacks controlled by the `Comm` object instead of embedding communication-related logic inside the Nbgl-related code.

This allows decoupling, and was essential to allow NBGL to work with either version of the `io` module.

This refactoring should be completely transparent to apps (except, it makes `init_comm` redundant).

## Case studies

The following PRs use the new io framework.
- [Boilerplate](https://github.com/LedgerHQ/app-boilerplate-rust/pull/164)
- [Vanadium](https://github.com/LedgerHQ/vanadium/pull/175)

(Vanadium is a lot more complex than the typical app in terms of communication complexity, and the PR includes substantial adaptations to take advantage of the increased IO buffer size.)


ledger_device_sdk/src/io_new.rs
```diff
@@ -0,0 +1,650 @@
+use crate::seph::{self, PacketTypes};
+
+pub use crate::io_legacy::{ApduHeader, Event, Reply, StatusWords};
+
+use crate::io_callbacks::nbgl_register_callbacks;
+
+#[cfg(any(
```
iartemov-ledger, line 7: No `nanosplus` ?

ledger_device_sdk/src/io_new.rs
```diff
@@ -0,0 +1,650 @@
...
+        0xa7 => {
+            let _ = comm.begin_tx().send(StatusWords::Ok);
+            crate::exit_app(0);
+        }
+        // Unknown INS within BOLOS namespace
+        _ => {
+            let _ = comm.begin_tx().send(StatusWords::BadIns);
+        }
+    }
+}
```
iartemov-ledger, line 650: Does it make sense to add some unit tests ?

iartemov-ledger (commented):
My five cents.

ledger_device_sdk/src/io_new.rs
bigspider: Copied from legacy, but `ItcUxEvent` seems to be only for BLE-enabled devices.

yogh333: @iartemov-ledger No because it is linked with Bluetooth

ledger_device_sdk/src/io_new.rs
```diff
@@ -0,0 +1,650 @@
...
+
+pub const DEFAULT_BUF_SIZE: usize = 273;
+
+pub struct Comm<const N: usize = DEFAULT_BUF_SIZE> {
+    buf: [u8; N],
+    expected_cla: Option<u8>,
+
+    apdu_type: u8,
+    #[cfg(not(any(target_os = "stax", target_os = "flex", target_os = "apex_p")))]
```
yogh333, line 36: To be replaced with `#[cfg(any(target_os = "nanosplus", target_os = "nanox"))]`

ledger_device_sdk/src/io_new.rs
```diff
@@ -0,0 +1,650 @@
...
+    pending_length: usize,
+}
+
+impl<const N: usize> Comm<N> {
+    pub fn new() -> Self {
+        let mut comm = Self {
+            buf: [0; N],
+            expected_cla: None,
+            apdu_type: PacketTypes::PacketTypeNone as u8,
+            #[cfg(not(any(target_os = "stax", target_os = "flex", target_os = "apex_p")))]
```
yogh333, line 52: Same as above

ledger_device_sdk/src/io_new.rs
```diff
@@ -0,0 +1,650 @@
...
+
+impl From<ApduError> for StatusWords {
+    fn from(e: ApduError) -> Self {
+        match e {
+            ApduError::BadLen => StatusWords::BadLen,
+        }
+    }
+}
+
+// TODO: we might not need to split DecodedEvent from DecodedEventType. Simplify.
```
yogh333, line 172: Agreed but simplification can be made in another future PR...

ledger_device_sdk/src/io_new.rs
```diff
@@ -0,0 +1,650 @@
...
+    }
+
+    pub fn reply<T: Into<Reply>>(self, data: &[u8], reply: T) -> Result<(), CommError> {
+        self.into_tx().extend(data)?.send(reply)?;
+        Ok(())
+    }
+}
+
+/// Immutable read view.
+pub struct Rx<'a, const N: usize = DEFAULT_BUF_SIZE> {
```
yogh333, line 440: Do we really need to make `Rx` public ? I mean, from the application point of view, it can be dangerous to manipulate the Rx/Tx object...

ledger_device_sdk/src/io_new.rs
```diff
@@ -0,0 +1,650 @@
...
+    pub fn consume(self) -> () {}
+
+    /// Decode into a higher-level event. No replies are sent, but UX-related and other OS interactions are dealt with.
+    pub fn decode_event(self) -> DecodedEvent<N> {
+        DecodedEvent::new(self.comm, self.len)
+    }
+}
+
+/// Mutable write view for building a send.
+pub struct Tx<'a, const N: usize = DEFAULT_BUF_SIZE> {
```
yogh333, line 468: Same comment as `Rx`

ledger_device_sdk/src/io_new.rs
yogh333, line 650: A more general comment: it could be relvant to split the module into submodule to ease readability and maintenance. The new io module is far better in terms of design and architecture but it is still a big file of several hundred of lines of code. WDYT ?

ledger_device_sdk/src/nbgl.rs
```diff
@@ -30,8 +31,6 @@ pub use nbgl_spinner::*;
 pub use nbgl_status::*;
 pub use nbgl_streaming_review::*;
 
-static mut COMM_REF: Option<&mut Comm> = None;
-
```
yogh333, line 34: This is great ! This was really ugly...

yogh333 (changes requested):
This is great, thank you @bigspider !

ledger_device_sdk/src/io_new.rs
bigspider: None in 8d3bec3512e3678a926280f0b289402aae534277.

At some point we should add better aliases for the various device groups/features, so this could become something similar to

```
#[cfg(has_buttons)]
```

ledger_device_sdk/src/io_new.rs
bigspider: Done in 8d3bec3512e3678a926280f0b289402aae534277.

ledger_device_sdk/src/io_new.rs
bigspider: As this is an object that apps would use, renamed to `CommandResponse` in 51e82242cc673d29f3b7334e647ae36ddcf558dc (and renamed `begin_tx` to `begin_response`).

ledger_device_sdk/src/io_new.rs
yogh333: @bigspider yes it is feasible, check for instance this PR: https://github.com/LedgerHQ/ledger-device-rust-sdk/pull/297 where there is an example for each NBGL use case.

bigspider: In #297 I see code examples, not automated tests.

Orchestrating integration tests (e.g. with speculos) as part of the CI for a mock app would be useful, but it's quite a bit of work and would be interesting for another PR.

yogh333: Yes agreed

bigspider:
> OK LGTM Please bump version for `ledger_device_sdk` and I will approve the PR

Since you're bumping the version in #297, we could merge this without bumping the version and bundle them in the same release - Vanadium will stay on [its own branch](https://github.com/LedgerHQ/ledger-device-rust-sdk/tree/vanadium) anyway until the issue with the custom compilation flags is addressed.

yogh333 (approved):

PR #297, opened 2025-09-23
Add support for new NBGL use cases

yogh333:


ledger_device_sdk/src/nbgl.rs
```diff
@@ -59,13 +81,32 @@ impl From<SyncNbgl> for u8 {
...
+static mut G_CONFIRM_ASK_WHEN_TRUE: bool = false;
+static mut G_CONFIRM_ASK_WHEN_FALSE: bool = false;
+static mut G_CONFIRM_MESSAGE_WHEN_TRUE: Option<CString> = None;
+static mut G_CONFIRM_SUBMESSAGE_WHEN_TRUE: Option<CString> = None;
+static mut G_CONFIRM_OK_TEXT_WHEN_TRUE: Option<CString> = None;
+static mut G_CONFIRM_KO_TEXT_WHEN_TRUE: Option<CString> = None;
+static mut G_CONFIRM_MESSAGE_WHEN_FALSE: Option<CString> = None;
+static mut G_CONFIRM_SUBMESSAGE_WHEN_FALSE: Option<CString> = None;
+static mut G_CONFIRM_OK_TEXT_WHEN_FALSE: Option<CString> = None;
+static mut G_CONFIRM_KO_TEXT_WHEN_FALSE: Option<CString> = None;
```
bigspider, line 104: Not a big fan of the statics here; each of them will occupy memory even when not used (the actual string content is in the heap, but the `Option<CString>` still takes at least 12 bytes for pointer, length and capacity, possibly more).

A solution might be a single enum with an entry for each of the possible screens, where each entry contains exactly the required fields for the currently shown screen. Putting the enum in a single `static mut G_SCREEN_STRINGS: Box<UiScreenStrings>` should address this, and also group together strings that are used together, which might be more maintainable.

(The current space occupation is maybe not yet problematic, so you might want to do that in a separate PR if you prefer.)

bigspider (approved):
LGTM, only some small comments.

ledger_device_sdk/README.md
```diff
@@ -1,14 +1,17 @@
...
 
-- some safe wrappers over common syscalls
-- IO abstractions
-- signature abstractions
-- UI libraries (the `ui` module for Nano (S/SP/X) apps, `nbgl` module for Stax and Flex apps)
+- Safe wrappers over common syscalls and C SDK functions
+- IO abstractions (`io` and `seph` modules)
+- Cryptographic abstractions (`ecc`, `hash` and `hmac` modules)
+- Aritmethic (simple and modular) abstraction (`math` module)
+- Persistent data storgae (`nvm` module)
```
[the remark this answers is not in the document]
yogh333: Done

ledger_device_sdk/src/nbgl.rs
yogh333: Yes good proposal, do you mean like this ?
```
pub(crate) enum ConfirmChoiceStrings {
    Undefined,
    WhenTrue {
        message: Option<CString>,
        submessage: Option<CString>,
        ok_text: Option<CString>,
        ko_text: Option<CString>,
    },
    WhenFalse {
        message: Option<CString>,
        submessage: Option<CString>,
        ok_text: Option<CString>,
        ko_text: Option<CString>,
    },
}
static mut G_CONFIRM_STRINGS: Box<ConfirmChoiceStrings> = Box::new(ConfirmChoiceStrings::Undefined);
```

PR #298, opened 2025-09-29
Y333/manage nbgl static strings

yogh333:


bigspider (approved):
Cleaner than before, although we might still want to look for a centralized container for _all_ NBGL use cases so they can share the same global space.

PR #371, opened 2026-03-10
Y333/port examples on io new

yogh333:
This PR ports all examples to `io_new` module


bigspider (approved):
LGTM, although if we want to publish this about at the same time as the migrated app-boilerplate-rust, then we should probably first make a new sdk release, so we can merge both this PR and [the boilerplate update PR](https://github.com/LedgerHQ/app-boilerplate-rust/pull/189) at the same time.

If we make a new release, maybe commit baf5f9e37afa03dcef9b0d6be2a0860a6f758519 should be first merged as a separate PR.

PR #378, opened 2026-03-16
Add stack usage APDUs handling

yogh333:


ledger_device_sdk/src/testing.rs
```diff
@@ -3,6 +3,166 @@ use core::panic::PanicInfo;
...
+/// 2. After the operation you want to measure, call [`StackTracker::get_usage`]
+///    to get the high-water mark in bytes.
+///
+/// # Safety
+/// These functions use inline assembly to read the current stack pointer (`sp`)
+/// and directly write to the stack memory region. They must only be called
+/// from the app's main execution context (not from interrupts).
+pub struct StackTracker;
+
+const STACK_INIT_VALUE: u8 = 0xFF;
```
bigspider: I think `0x00` and `0xFF` should be avoided as marker values, as they are more likely to be written by unrelated code (increasing the risk of making the measurement less precise).

A more random looking constant like `0xAC` (which I chose because it's the stACk :P) would be better

ledger_device_sdk/src/testing.rs
```diff
@@ -3,6 +3,166 @@ use core::panic::PanicInfo;
...
+    unsafe static _stack: u8;
+    unsafe static _estack: u8;
+}
+
+impl StackTracker {
+    /// Paint the unused portion of the application stack with a known pattern.
+    ///
+    /// This writes `0xFF` from `_stack` (bottom of stack region) up to the
+    /// current stack pointer, preserving the currently active stack frames.
+    /// The canary (just below `_stack`) is not touched.
```
bigspider, line 37: Comments here and below should probable use the name `STACK_INIT_VALUE` instead of its value.

bigspider (approved):
Nits.

ledger_device_sdk/src/testing.rs
yogh333: Done

ledger_device_sdk/src/testing.rs
yogh333: Done

PR #381, opened 2026-04-17
fix deadlock in io new and io legacy

mbrousset-ledger:
Fixing cases where speculos tests block, when the test ends with a status and receives CLA=B0 INS=57 before it ends (typically pytest sequence with `--get-stack-consumption`)


yogh333 (changes requested):
please put those modifications under `#[cfg(feature = "stack_usage")]` 
thank you !

yogh333 (approved):

PR #404, opened 2026-06-15
Add workflow to manage nightly bump in all Rust apps

yogh333:


yogh333:
Not needed as Rust toolchain is overrided by `RUSTUP_TOOLCHAIN` env variable

PR #408, opened 2026-06-26
Enable ML-KEM and ML-DSA for rust applications

srasoamiaramanana-ledger:
This PR adds the Rust API for ML-KEM and ML-DSA C SDK functions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,6 +98,8 @@ jobs:
         working-directory: ledger_device_sdk
         env:
           LEDGER_SDK_PATH: /tmp/c_sdk
+          # Shrink the heap for Nano X only, other targets keep the default.
+          HEAP_SIZE: "nanox: 2048"
         run: |
           cargo test --target ${{ matrix.target }} --features unit_test,mlkem,mldsa --tests
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,6 +80,27 @@ jobs:
         run: |
           cargo test --target ${{ matrix.target }} --features unit_test --tests
 
+  test-pqc:
+    name: Run ML-KEM and ML-DSA unit tests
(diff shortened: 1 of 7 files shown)


yogh333 (commented):
Thank you for the PR!
Is there a way to make unit tests run (dedicated branch of C SDK)?
If yes, then `ci.yml` workflow should be updated accordingly by enabling the features and pointing to the proper C SDK branch (using `LEDGER_SDK_PATH` env) variable.

.github/workflows/ci.yml
```diff
@@ -72,13 +72,22 @@ jobs:
...
+          - { name: "mldsa", features: "unit_test,mldsa", c_sdk_branch: "sdk-mldsa-support" }
     steps:
       - name: Clone
         uses: actions/checkout@v6
-      - name: Unit tests
+      - name: Setup C SDK (${{ matrix.variant.name }})
+        if: ${{ matrix.variant.c_sdk_branch != '' }}
+        run: |
+          git clone https://github.com/LedgerHQ/ledger-secure-sdk --branch ${{ matrix.variant.c_sdk_branch }} --single-branch /tmp/c_sdk
+          echo "LEDGER_SDK_PATH=/tmp/c_sdk" >> $GITHUB_ENV
```
yogh333: Should be `export LEDGER_SDK_PATH=/tmp/c_sdk`

srasoamiaramanana-ledger: Ok :+1: 
The CI fails on the C SDK branches because of the API level which is not set. In my opinion it's better to first merge the C SDK branches, cherry-pick on API level 26 then update the ci.yml. What do you think?

kmeudic-ledger (approved):

.github/workflows/ci.yml
yogh333: Yes agreed

yogh333 (approved):

PR #418, opened 2026-08-03
Merge develop into master (publish)

mbrousset-ledger:


overcat:
It looks like the latest Docker image update ([commit 5abb19e](https://github.com/LedgerHQ/ledger-app-builder/commit/5abb19ed2e0b23acd8c1818ca6e019cd90863246)) has introduced a breaking change that is affecting our build pipeline.

Could you please share if there is an estimated timeline for the next SDK release?

Thanks for your help!

